PR #6, opened 2023-03-02
fix: rename image build

castrojo:


castrojo:
https://github.com/ublue-os/main/pull/7 fixes this better

PR #16, opened 2023-03-03
feat: Add kernel-devel package

joshua-stone:
The kernel-devel package is important for building out-of-tree modules and has a chance of failing to install on an update if layered. Having this package in the base image will ensure that downstream images and users don't get this update issue.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1,7 +1,9 @@
 {
     "all": {
         "include": {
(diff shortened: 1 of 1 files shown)


marcoceppi (approved):

castrojo (approved):
Worth it, only 11MB!

PR #19, opened 2023-03-03
feat: install libva, and nvtop

joshua-stone:


castrojo (approved):

castrojo (approved):
I don't think we should block on 38 being broken do we?

bigpod98 (approved):

joshua-stone:
I'm going to defer the inclusion of ffmpeg and mesa-va-drivers to another change and add other packages in the meantime.

PR #43, opened 2023-03-04
docs: add a link to nvidia repo for those users

castrojo:


atimeofday (approved):
Good tweak for clarity, good redirect to keep things tied together for users while keeping repos loosely coupled and self-sufficient

marcoceppi (approved):

PR #87, opened 2023-03-24
feat: new akmods build process

dhoell:
See my PRs in -config, -akmods, -nvidia. 

- No technical changes required
- Build time adjusted to be more in line with quay.io/fedora-ostree-desktops, this should fix common occurrences where the fedora repositories have been updated before we build the images, leading to mismatched packages. Build times in other repos have been adjusted accordingly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,19 +1,10 @@
 name: build-ublue
 on:
   pull_request:
-    branches:
-      - main
-    paths-ignore:
-      - '**.md'
-      - '**.txt'
+  merge_group:    
   schedule:
     - cron: '0 7 * * *'  # 7 am everyday
-  push:
-    branches:
-      - main
-    paths-ignore:
-      - '**.md'
-      - '**.txt'
+  workflow_dispatch:
 env:
     IMAGE_BASE_NAME: main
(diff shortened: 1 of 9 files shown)


castrojo (approved):

.github/workflows/build.yml
```diff
@@ -7,7 +7,7 @@ on:
       - '**.md'
       - '**.txt'
   schedule:
-    - cron: '20 20 * * *'  # 8:20pm everyday
+    - cron: '0 7 * * *'  # 7 am everyday
```
bsherman, line 10: Question on this change... this is 7:00am UTC.

I checked the last several days of https://quay.io/repository/fedora-ostree-desktops/silverblue?tab=tags (also looked at base, kinoite, sericea, vauxite) and it seems they are generally done building by 4am UTC...

So it seems this is a good change as it leaves a few hours of buffer for those to complete. 

But I wanted to confirm as the original 8:20pm time was also planned to run an hour or so after a daily build. What changed? Was the old time based on the cgwalters images?

bsherman: This was answered in Discord. New time is as I expected. Looks good to me.

bsherman (approved):

castrojo (approved):

PR #139, opened 2023-04-11
feat: Include libratbag

dhoell:
Reason for inclusion: Package not available from flathub; hardware enablement.

Add `libratbag-ratbagd` (https://github.com/libratbag/libratbag) to packages, required to configure various mice. A GUI (https://github.com/libratbag/piper) is available from flathub and does not need to be installed in the image, however without ratbagd it is nonfunctional.

Libratbag does not expose any GUI programs to the user and is "invisible" when installed. User which require its functionality are free to install piper from flathub. Its installed size amounts to 336.79 KB.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ You should be familiar with [immutable desktops](https://silverblue.fedoraprojec
   - Hardware acceleration and codecs
   - `distrobox` for terminal CLI and user package installation
   - A selection of [udev rules and service units](https://github.com/ublue-os/config)
+  - [libratbag](https://github.com/libratbag/libratbag), to configure supported mice via [piper](https://github.com/libratbag/piper)
   - Various other tools: check out the [complete list of packages](packages.json)
 - Sets automatic staging of updates for the system
 - Sets flatpaks to update twice a day
--- a/packages.json
+++ b/packages.json
@@ -12,6 +12,7 @@
                 "intel-media-driver",
                 "just",
                 "kernel-devel-matched",
+                "libratbag-ratbagd"
                 "libva-intel-driver",
                 "libva-utils",
                 "mesa-va-drivers-freeworld",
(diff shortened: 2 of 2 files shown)


dhoell:
Not from a normal one, maybe from a rootful one, let me check.
But if we include things like `htop`, `vim`, and `fzf` we can definitely include this.

joshua-stone:
This looks good.

In the future we may also want to investigate adding [OpenRazer](https://openrazer.github.io/) which uses DKMS.

dhoell:
IMO: 
- If it's available as a flatpak: MUST NOT be included.
- If it works in a distrobox: SHOULD NOT be included (exceptions apply, e.g. `vim`,  for editing /etc files on the host)
- If it only works on the host (mostly DE and driver stuff): MAY be included.

As most of the "hardware enablement" packages, aka drivers and adjacent stuff like this one, are pretty small I would be in favor of a broad inclusion policy. Most users understandably don't want to muck about to get their gamepads/mice/laptops/etc working properly. 

Some lines will obviously have to be drawn (and have been drawn, see: separate images for nvidia, inclusion of intel drivers in main, even though AMD (i)GPUs exist,...). To be clear, I think those decisions were correct. While not as widespread as Intel iGPUs, I'd say that mice configurable by libratbag are widespread enough to justify inclusion.

castrojo:
> Otherwise how is the line drawn with respect to what should be included in main?

We've started to codify that here so that we're more explicit about the desire to cover the hardware use cases: https://github.com/orgs/ublue-os/discussions/125 (Feedback there would be welcomed!) 

I think a good measuring stick for this one is "if you plug this device into a windows or mac machine it works" but if you install it on one of our images it doesn't work. We want to avoid users having to layer for hardware support.

castrojo (approved):

PR #151, opened 2023-04-16
fix: update image-name parameter

joshua-stone:


joshua-stone:
Closing since this is a redundant PR.

PR #153, opened 2023-04-16
refactor: avoid subshell, sign image with env. var

SimonWoodtli:
* The private key for image signing does not need to be stored in the building process. But can be directly accessed from the env. variable.
* Use parameter expansion to store the short sha commit.


akdev1l (approved):

castrojo:
Thanks for this, we'll get to this after F38 and we have some green on the board. 😄

SimonWoodtli:
You're welcome, yeah F38 has third party repo issues right? Anyway now it only fails in CI and no longer on our machines 😄

PR #177, opened 2023-04-25
feat: switch the latest tag to F38

joshua-stone:
F38 builds seem to have stabilized since the release date last week, and many upgrade bugs should be fixed by now.


bsherman (approved):
Make it so.

PR #190, opened 2023-04-30
feat: add tmux

bsherman:
per some chat in discord, this would be nice in main


akdev1l (approved):
I already add this in my image so +1 from me

PR #202, opened 2023-05-14
feat: add alsa-tools package

### LINKED ISSUE #201: Alsa-tools incorporated into all ISOs
### Describe the package

Alsa-tools package is useful for configuring non-working sound cards & speaker setups (hdajackretask). F.e. most built-in Realtek soundcards don't have front 3.5mm panel recognized out of the box with Pipewire. Some laptops have multiple speaker sound system which are either recognized as mono, or not recognized at all. Also useful for troubleshooting other sound related issues.

Since those individual sound fixes are much harder to incorporate into an ISO, it's at least useful to have the package out of the box, which is required for those fixes from the end user.

### Image

All Images

castrojo:
Fixes #201

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -2,6 +2,7 @@
     "all": {
         "include": {
             "all": [
+                "alsa-tools",
                 "apr",
                 "apr-util",
                 "distrobox",


Arcitec (approved):
Claaaaw! This is indeed really useful for people with those soundcards, and the binaries are tiny, so let's bundle them! :)

Arcitec (changes requested):
Fiftydinar brought up that those tools also install desktop shortcuts. Perhaps we can do a post-install script step to delete those `.desktop` files.

castrojo:
Ok setting this to draft while we investigate pipewire solutions.

Arcitec:
@castrojo We can close this. It turned out that alsa-tools try to write to immutable locations, so we can't add this anyway.

More importantly, the issue is solvable with pipewire/wireplumber for sure (since alsa-tools just talks directly to pipewire's ALSA emulation), so the fact that the devices show up in alsa-tools means that Pipewire is able to see them. It would just require some investigation by OP to find the correct config for their system. :)

PR #212, opened 2023-05-17
chore: add new public key for akmods

castrojo:


bsherman (approved):
I'm happy to see this key has a good Issuer:

`Issuer: O = Universal Blue, OU = akmods`

PR #251, opened 2023-06-20
feat: add GPU computer packages 

EliasOfWaffle:
Add GPU computer packages
Intel OpenCL
Intel Level Zero 
Mesa OpenCL (Clover and RustICL)


castrojo:
Can we open an issue along side this? We need to start thinking about ARM builds and we might also want to consider some sort of other parameters, also would like to know the disk size impact of these, thanks!

EliasOfWaffle:
> Can we open an issue along side this? We need to start thinking about ARM builds and we might also want to consider some sort of other parameters, also would like to know the disk size impact of these, thanks!

yes, I'm sorry, I didn't realize about the other architectures, that would be a great issue, thanks

PR #262, opened 2023-06-25
feat: Fix inconsistent thumbnails across images

EyeCantCU:
Addresses an inconsistency where other images can view thumbnails just fine in their respective file managers, but others can't

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -79,6 +79,7 @@
                 "openbox",
                 "pcmanfm-qt",
                 "perl-File-MimeInfo",
+                "qt5-qtimageformats",
                 "qterminal",
                 "sddm",
                 "sddm-themes",
--- a/packages.json
+++ b/packages.json
@@ -44,7 +44,8 @@
             "silverblue": [
                 "adw-gtk3-theme",
                 "ffmpegthumbnailer",
-                "gnome-tweaks"
+                "gnome-tweaks",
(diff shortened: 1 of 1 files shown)


fiftydinar:
GTK based DEs work too (tested), so add these too to the list.

Cinnamon doesn't have those in repos, so they need to be built manually.

castrojo:
What parts need to be built for Cinnamon?

fiftydinar:
> What parts need to be built for Cinnamon?

ffmpegthumbnailer & raw-thumbnailer.

For LxQT:

ffmpegthumbnailer & qt5-imageformats

castrojo:
We should just be able to add them to Cinnamon here right? The two binaries are in the fedora repos already:  https://github.com/ublue-os/cinnamon

fiftydinar:
> We should just be able to add them to Cinnamon here right? The two binaries are in the fedora repos already: https://github.com/ublue-os/cinnamon

Yup.

Weirdly, I did not get those packages out of the box, and when I tried to install them in terminal from live-installation, it reported that those packages are not avaliable.

EyeCantCU:
All right. I'll get this updated. Thank you for the help with this

EyeCantCU:
Opening a PR for Cinnamon as well since it isn't built here

castrojo (approved):

PR #274, opened 2023-07-16
feat: add power-profiles-daemon

bubuntux:
adding [power-profiles-daemon](https://packages.fedoraproject.org/pkgs/power-profiles-daemon/power-profiles-daemon/fedora-38-updates.html#dependencies) to all versions, power profile allows setting [cpu frequency scaling](https://wiki.archlinux.org/title/CPU_frequency_scaling) , gnome and kde already provide a GUI to change this setting (once this package is installed) for other versions, powerprofilesctl command can be used as need it, this is specially useful for laptops since allow you to lower the cpu frequency to save battery or change to high performance mode to get the most of the de laptop when pluged-in.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -33,6 +33,7 @@
                 "openrgb-udev-rules",
                 "openssl",
                 "podman-compose",
+                "power-profiles-daemon",
                 "pipewire-codec-aptx",
                 "smartmontools",
                 "symlinks",


bigpod98:
Does this interfere with how gnome setings do this

bubuntux:
my understanding is that gnome uses this to change that setting, let me do a quick test to verify....

bubuntux:
i rebased into silverblue and the package is already there lol, like i was suspecting this is already used by gnome and kde

gerblesh:
it's already included by default, best not to manually install it. Both KDE and GNOME use it by default, using their own GUI to control it through the dbus interface.

bubuntux:
```
it's already included by default, best not to manually install it. Both KDE and GNOME use it by default, using their own GUI to control it through the dbus interface.
```

right, but other versions like sericea is not installed by default and comes with nothing similar, if we set it for all it will not change gnome/kde since it's already there

gerblesh:
> ```
> it's already included by default, best not to manually install it. Both KDE and GNOME use it by default, using their own GUI to control it through the dbus interface.
> ```
> 
> right, but other versions like sericea is not installed by default and comes with nothing similar, if we set it for all it will not change gnome/kde since it's already there

It's best to stick to the defaults of a main image, there's no way to control power profiles daemon from the stock sway spin, if one wants power management then they can set it up themselves in sericea. Or Fedora can do it upstream.

bubuntux:
yeah you are probably right

PR #300, opened 2023-08-13
chore(ci): Enable PR tagged images on pull request approval

EyeCantCU:
Submits an image to the GitHub Container Registry on pull request approval, permitting approved pull requests to be tested before being merged.

Requires branch protection rule: 'Require approval of the most recent reviewable push'

This rule ensures that the state of a PR is reset after a new commit has been pushed to an open pull request.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,7 +1,9 @@
 name: build-ublue
 on:
   pull_request:
-  merge_group:    
+  pull_request_review:
+    type: [submitted]
+  merge_group:
   schedule:
     - cron: '0 7 * * *'  # 7 am everyday
   workflow_dispatch:
@@ -12,6 +14,7 @@ env:
 jobs:
   push-ghcr:
     name: Build and push image
+    if: github.event.review.state == 'approved' || github.event_name != 'pull_request_review'
     runs-on: ubuntu-22.04
     permissions:
       contents: read
@@ -34,7 +37,7 @@ jobs:
           # When F38 is added, sericea will automatically be built too
           - image_name: sericea
             major_version: 37
-    steps: 
+    steps:
       # Checkout push-to-registry action GitHub repository
       - name: Checkout Push to Registry action
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


KyleGospo:
Love the concept, only thing is we need to make sure that commits made after a PR is approved for runs aren't run until re-approved, otherwise you could get a PR approved and then do something malicious after the fact.

EyeCantCU:
Very true. I'll look into it

EyeCantCU:
So, here's what I'm thinking. Instead of using the `pull_request` trigger, we could use the `pull_request_review trigger` on submissions. When a pull request is approved, the workflow will run. To avoid future pushes from triggering the build workflow again, we'd need to enable the branch protection rule: 'Require approval of the most recent reviewable push'. For each action in the workflow, we'd need an if statement that checks for whether something was pushed versus the state of the PR being approved. Thoughts?

EyeCantCU:
The immediate issue i see here is contributors using PRs for revisions aren't immediately going to see build results until workflows are approved. Luckily I think restoring the PR trigger should resolve that

EyeCantCU:
Everything is in place now for this to work properly, save for the branch protection rule

bsherman:
> The immediate issue i see here is contributors using PRs for revisions aren't immediately going to see build results until workflows are approved. Luckily I think restoring the PR trigger should resolve that

This _hopefully_ isn't a big problem. Hmm... except maybe? Contributors should be encouraged to run their builds locally... but I guess workflow changes are something not as easily tested since i don't think they'll run in a fork.

EyeCantCU:
> > The immediate issue i see here is contributors using PRs for revisions aren't immediately going to see build results until workflows are approved. Luckily I think restoring the PR trigger should resolve that
> 
> This _hopefully_ isn't a big problem. Hmm... except maybe? Contributors should be encouraged to run their builds locally... but I guess workflow changes are something not as easily tested since i don't think they'll run in a fork.

My latest push fixed this

EyeCantCU:
Latest push ensures that the PR number is applied to the tag and that builds triggered by pull_request_review only run on approvals

PR #301, opened 2023-08-13
feat: enable fedora 39 builds

p5:
A bit of an experiment to see how much needs changing for us to start building against Fedora 39. 
The images are now being pushed to Quay (or at least for silverblue).


yodatak:
Interedted because it could be easy to debug n+1 fedora

EyeCantCU (approved):
Woohoo!

.github/workflows/build.yml
```diff
@@ -21,29 +21,37 @@ jobs:
...
           - major_version: 37
             is_latest_version: false
             is_stable_version: true
+            is_gts_version: false
           - major_version: 38
             is_latest_version: true
             is_stable_version: true
+            is_gts_version: true
+          - major_version: 39
+            is_latest_version: true
```
p5, line 36: I can't remember the decision - are we marking beta releases as the latest?  

I seem to think we wanted latest to be the latest stable release

EyeCantCU: Latest and stable are required to push to the latest tag so the latest stable release is what gets pushed there. 39 won't get pushed to that until it is marked as stable. When that happens, we set the flag for latest and stable releases to false for 38 but keep gts true and that will be the release that stays behind. Setting the latest flag here for 39 won't have an impact on our tags as it isn't marked stable yet. I know that sounds confusing too

p5: Thanks!  Happy for this to be merged then

EyeCantCU: Awesome! Let's get this rolled out :)

castrojo: Nice work y'all!!

szescxz:
@EyeCantCU 
While I only tested on Bazzite, I believe `google-noto-sans-cjk-fonts` is still needed on F39 builds.
More info here: https://github.com/ublue-os/bazzite/issues/153#issuecomment-1801432250

KyleGospo:
Original fix is restored on Bazzite.

EyeCantCU:
> @EyeCantCU 
> While I only tested on Bazzite, I believe `google-noto-sans-cjk-fonts` is still needed on F39 builds.
> More info here: https://github.com/ublue-os/bazzite/issues/153#issuecomment-1801432250

Thank you for letting me know! I'll see about getting it restored in main momentarily

PR #303, opened 2023-08-16
chore(ci): Bump isogenerator to v2.1.0

EyeCantCU:
This bumps isogenerator to v2.1.0

### DIFF
--- a/.github/workflows/build.yml.save
+++ b/.github/workflows/build.yml.save
@@ -1,0 +1,192 @@
+name: build-ublue
+on:
+  pull_request:
+  pull_request_review:
+    type: [submitted]
+  merge_group:
+  schedule:
+    - cron: '0 7 * * *'  # 7 am everyday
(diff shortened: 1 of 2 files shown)


KyleGospo (approved):

castrojo:
Please revert, we do this on purpose, I can explain when I'm not on mobile

castrojo:
Or rather, pin the explicit version!

EyeCantCU:
All right. I'll fix this up

castrojo:
For some context, this is so when if someone is revving on the isogenerator to test in other images we can always ensure that main/nvidia are releaseable on a known-good version. 

Then we explicitly bump after testing. Thanks!

KyleGospo (approved):

EyeCantCU:
That makes a lot of sense. Thank you for explaining that to me. Wouldn't want to ship a bad ISO

PR #325, opened 2023-08-31
feat: use artifacts to streamline image builds

bsherman:
The goal of this PR is to streamline image builds.

The idea is to split `*-main` and `*-nvidia` image builds into 2 jobs.
The `*-main` job will run first, and each matrix job instance will build the `foo-main` image, but there will be a new step, the image will be `podman save`d to a tarball and uploaded as an artifact.
The `*-nvidia` job will start after all `*-main` instances finish. This job will have a new step of downloading the previously uploaded artifact, and then  `podman load`ing the tarball so that the image build can continue.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ env:
     IMAGE_REGISTRY: ghcr.io/${{ github.repository_owner }}
 
 jobs:
-  push-ghcr:
+  push-main:
     name: Build and push image
     runs-on: ubuntu-22.04
     permissions:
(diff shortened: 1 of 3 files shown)


bsherman:
This approach is not going to work for main/nvidia since the resulting artifacts are only used once and are so large that the upload process takes a long time.

It MAY still be a viable solution for `akmods` repo where we want a single "ready to build" image which all kmods can be built against.

bigpod98:
What is purpose of using artifacts when you are doing essentially same thing as pulling from registry anyway

bsherman:
> What is purpose of using artifacts when you are doing essentially same thing as pulling from registry anyway

There's a few ideas which resulted in this trial.  One is how to have distinct images built where the second image depends on the first but NOT use a registry... this is important for multi-job builds (like this trial demonstrated) since we don't push to registry during PR builds.

bsherman:
This is a defunct approach. See https://github.com/ublue-os/main/issues/356

PR #326, opened 2023-08-31
feat: split main/nvidia into distinct steps within job

bsherman:
The goal of this PR is to streamline image builds.

**Current problem:**
both `*-main` and `*-nvidia` images are built once each for every permutation of a matrix comprised of: fedora major version (37, 38), nvidia driver version (0 [signifies main, no nvidia], 470, 535), flavor (main or nvidia), and variant (silverblue, kinoite, FOO, etc).

Clarification: buildah action is smart enough, that by using the `target` argument, it does not attempt building bogus nvidia images for `FOO-main` with nvidia verison `0` However, for each build job that results only in a pushed `FOO-nvidia` image, there is always a full `FOO-main` image built as part of that process. 

**Ideal Solution:**
The most ideal solution should be to build `*-main` only once in a first job (matrix of fedora version and variant), then build `*-nvidia` once for each nvidia version in a second job. However, this approach fails because:
1. we don't push to registries in PRs
2. due to the large size of images, both github artifacts are unusable as temporary storage (see: #325)

**This PR's approach:**
This PR as implemented keeps the builds in a single job, but splits the Containerfile into `main` and `nvidia` and builds them in sequential steps.

This is still not perfect, as while this removes the standalone `FOO-main` with nvidia version `0` build, which does cut out 13 distinct matrix jobs (given today's version matrix), each running job does still build `FOO-main` and even publishes that image once from each job.

Some other detail about the implementation below:
There will now be both a main and nvidia version of each of these steps:
- Generate Tags
- Image Metadata
- Build Image
- Push to GHCR
- Sign container image

There is some duplication in workflow build.yml, but it allows the use of distinct Containerfiles for each image, `main` and `nvidia`, and does not require completely building both images in distinct jobs as the previous matrix for `image_flavor` was doing.


bsherman:
This is a defunct approach. See https://github.com/ublue-os/main/issues/356

PR #332, opened 2023-09-12
fix: remove libheif-freeworld from packages to fix build

bsherman:
This is probably temporary until rpmfusion sorts things out but this is to fix all Fedora 38 builds failing with:

```
    error: Could not depsolve transaction; 1 problem detected:
     Problem: conflicting requests
      - package libheif-freeworld-1.15.1-4.fc38.x86_64 from rpmfusion-free requires libheif(x86-64) = 1.15.1, but none of the providers can be installed
      - package libheif-freeworld-1.16.1-1.fc38.x86_64 from rpmfusion-free-updates requires libheif(x86-64) = 1.16.1, but none of the providers can be installed
      - cannot install both libheif-1.15.1-2.fc38.x86_64 from fedora and libheif-1.16.2-1.fc38.x86_64 from @System
      - cannot install both libheif-1.16.1-1.fc38.x86_64 from updates-archive and libheif-1.16.2-1.fc38.x86_64 from @System

```

Also changed the default Containerfile build arg for FEDORA_MAJOR_VERSION to 38, since that's what should be default these days.

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -1,12 +1,12 @@
 ARG IMAGE_NAME="${IMAGE_NAME:-silverblue}"
 ARG SOURCE_IMAGE="${SOURCE_IMAGE:-silverblue}"
 ARG BASE_IMAGE="quay.io/fedora-ostree-desktops/${SOURCE_IMAGE}"
-ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION:-37}"
+ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION:-38}"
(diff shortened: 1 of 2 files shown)


bsherman:
@fiftydinar is verifying if this is a workable long term solution or needs to be temporary.

fiftydinar:
I am good with this to go before my test

If libheif fixed itself after some time, we can revert other related freeworld packages too

KyleGospo (approved):

fiftydinar:
Libheif is still broken, so we will wait till freeworld is fixed.

PR #348, opened 2023-09-19
feat: enable onyx builds

### LINKED ISSUE #196: Turn on onyx builds
Onyx (Budgie) is now upstream, we should turn this on when F39 builds start:

https://quay.io/repository/fedora-ostree-desktops/onyx

bobslept:
This should enable onyx builds for F39 en exclude them from F37 and F38

Closes #196


bigpod98 (approved):
This looks good

PR #350, opened 2023-09-22
feat: Add WL driver by default (Disabled by default)

KyleGospo:


bsherman (approved):
This looks good. Has already been tested out in bazzite, I believe.

szescxz:
I do get
```
libkmod: ERROR libkmod/libkmod-config.c:712 kmod_config_parse: /etc/modprobe.d/default-disable-broadcom-wl.conf line 3: ignoring bad line starting with 'wl'
```
whenever I try to `modprobe` stuff. Doesn't seem affecting anything but is this the intended behavior?

bsherman:
> _Because I assumed this should be `blacklist wl`_

Thank you! I obviously missed this in my review.

Issue #356, opened 2023-09-22
Reorganize main builds to better support image flavors: main, nokmods, nvidia, asus, surface


bsherman:
We previously completed #313 which merged `*-nvidia` image workflow into this `main` repo. But since then, we've noticed some excessive build times and duplication of builds due to the chained nature of main+nvidia builds. Additionally, we have some new build workflow needs as we are trying to support various laptop types, specifically: asus, framework, and surface.

To better handle the new situation we plan to do the following:
1. restore the `nvidia` repo and once again build `*-nvidia` images there, removing the `*-nvidia` complexity from `main` workflow (the nvidia kmods will still be built in `akmods` repo)
2. modify `main` repo workflow adding a `*-nokmods` image flavor for all variants (silverblue, kinoite, etc) which will not include any kmods from our `akmods` repo
3. `asus` and `surface` downstream images will start building `FROM ghcr.io/ublue-os/VARIANT-nomods` which will allow them to use custom kernel and kmods.
4. https://github.com/ublue-os/akmods/issues/68

End states:
1. Effectively `*-nokmods` will be the foundational image from which `*-main`, `*-asus`, and `*-surface` are built. 
2. Downstream image builds currently using `*-main` will not see a change (including `*-nvidia`, `bluefin` and `bazzite`)

bsherman:
Welp. That was quick.

This looks to be wrapped up except for point 4 which has it's own ticket: https://github.com/ublue-os/akmods/issues/68

@castrojo @KyleGospo @EyeCantCU

PR #357, opened 2023-09-22
chore(ci): use a single rpmfusion mirror for builds (temp disable F39)

bsherman:
This change selects a specific rpmfusion baseurl rather than using the metalink. During CI builds this will provide more consistent success/fail behavior over the current situation where the mirror metalink is used. This change only impacts the actual build time use in this repo; it is not propagated to user installed images.

Temporarily DISABLES F39 builds in workflow.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,6 +21,7 @@ jobs:
       matrix:
         image_flavor: [main, nvidia]
         image_name: [silverblue, kinoite, vauxite, sericea, base, lxqt, mate, onyx]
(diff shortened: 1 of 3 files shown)


bsherman:
Don't hate me for disabling F39... I just need to get this reorg done ( https://github.com/ublue-os/main/issues/356 ) and current F39 rpmfusion failures are blocking.

bsherman:
Expects https://github.com/ublue-os/akmods/pull/69 to be merged before this merges.

castrojo (approved):

PR #371, opened 2023-09-28
feat: Add OpenTabletDriver udev rules

KyleGospo:
Rather than build our own RPM, this grabs the portable tar.gz release straight from GitHub and pulls the rules out of it.
Needs further discussion, let's make sure we actually want to go this route. It's simpler than packaging our own RPM, but definitely not as clean.

### DIFF
--- a/nokmods-install.sh
+++ b/nokmods-install.sh
@@ -53,5 +53,17 @@ fi
 ## install packages direct from github
 /tmp/github-release-install.sh sigstore/cosign x86_64
 
+# Install OpenTabletDriver udev rules from their portable releases
+mkdir -p /tmp/OpenTabletDriver/
+mkdir -p /usr/etc/udev/rules.d/
+curl -s https://api.github.com/repos/OpenTabletDriver/OpenTabletDriver/releases/latest \
+| grep "browser_download_url.*opentabletdriver-.*-x64.tar.gz" \
+| cut -d : -f 2,3 \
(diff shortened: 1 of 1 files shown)


bsherman (changes requested):
I disagree with this approach because it's incredibly similar to what we already do in the config repo to build  our `ublue-os-udev-rules` RPM.

see: https://github.com/ublue-os/config/blob/main/Containerfile#L5

I think we should follow the same pattern, even if there's an added step needed to filter out non-udev stuff.

bsherman:
This and recent udev stuff makes me think we should have an agreed upon process for adding udev rules our different contributors alway add in a consistent way.

I'm imagining something that fits with what we already do, like ...

How to add new udev rules, order of preference:
1. use upstream RPM if available; add as `packages.json` "include" in `main`
2. use upstream spec if available, build RPM in COPR; add as `packages.json` "include" in `main`
3. use tarball of upstream git repo if fully dedicated to udev rules; add to `confing`repo and build as part of `ublue-os-udev-rules`
4. use tarball of upstream git repo, filter to only udev rules(and license, readme, etc); add to `confing`repo and build as part of `ublue-os-udev-rules`
5. copy udev rule files directly into `config` repo and build as part of `ublue-os-udev-rules


I think today we already have examples of steps 2, 3, and 5... this PR would be an example of 4.

PR #373, opened 2023-09-28
chore: remove podman workaround

SoMuchForSubtlety:
This has been fixed in podman 4.7.0 https://github.com/containers/podman/issues/19930#issuecomment-1737456096

(draft until it hits stable https://src.fedoraproject.org/rpms/podman)


castrojo:
Yeah and also we need to account for the delay between it being published in fedora and making it to the oci image that it's published so we don't remove it too early.

castrojo (approved):

bsherman:
@castrojo you beat me to this :-)

bsherman:
I've got this patch installed on my machine and confirmed it's working as expected.

PR #377, opened 2023-09-30
chore(podman): Remove pinned crun version

john-ghatas:
CRUN was pinned to 1.8.7 in the Containerfile a while back as a workaround for https://github.com/containers/podman/issues/19930. The issue is closed now with the release of podman 4.7.0, therefore the version can be unpinned.


john-ghatas:
Fedora 37 didn't receive the 4.7.0 update, but it will go eol in 2 and half months. Imo not worth it holding this back for that reason.

castrojo:
Covered with this, thanks anyway!

https://github.com/ublue-os/main/pull/373

PR #380, opened 2023-10-04
feat: add a text editor and image viewer to Silverblue

ArtikusHG:
By default ```silverblue-main``` does not come with a text editor and an image viewer. While I agree with the approach of main being a reasonably minimal base, not adding too much compared to the upstream image, these two apps provide core desktop functionality, very important for the majority of users. They are also a very small addition - 0.5 MB for ```gnome-text-editor```, 1.4 MB for ```eog``` ("eye of gnome", the image viewer for GNOME versions <45) and 2.1 MB for ```loupe``` (the image viewer for 45+).

Personally, not having an image viewer and text editor by default was one of the biggest annoyances for me after first installing ublue - I was really expecting a more complete GNOME experience. So, I think adding these apps is a great way to make using ublue a bit more convenient.

If anyone has questions/concerns I'm open to discussion.

### DIFF
--- a/nokmods-packages.json
+++ b/nokmods-packages.json
@@ -45,7 +45,8 @@
             ],
             "silverblue": [
                 "adw-gtk3-theme",
-                "gnome-tweaks"
+                "gnome-tweaks",
+                "gnome-text-editor"
             ],
             "kinoite": [
                 "icoutils",
@@ -225,7 +226,8 @@
                 "google-noto-sans-cjk-fonts"
             ],
             "silverblue": [
-                "raw-thumbnailer"
+                "raw-thumbnailer",
+                "eog"
             ],
             "mate": [
                 "f37-backgrounds-base",
@@ -248,7 +250,8 @@
                 "google-noto-sans-cjk-fonts"
             ],
             "silverblue": [
-                "raw-thumbnailer"
+                "raw-thumbnailer",
+                "eog"
             ],
             "mate": [
                 "f38-backgrounds-base",
@@ -273,6 +276,9 @@
             "all": [],
             "kinoite": [
                 "xwaylandvideobridge"
+            ],
+            "silverblue": [
+                "loupe"
             ]
         },
         "exclude": {


castrojo:
I think the intent they want is for those apps to be installed as flatpaks?

p5:
> I think the intent they want is for those apps to be installed as flatpaks?

Maybe an extremely limited (and unopinionated) implementation of the stuff added to Bluefin to install default apps could be good here?

ArtikusHG:
> I think the intent they want is for those apps to be installed as flatpaks?

We can implement that as well, we do have yafti after all :P

> Maybe an unopinionated implementation of the stuff added to Bluefin & Bazzite to install default flatpaks could be good here?
> Basic things like Contacts, Text Editor, Image viewer and the disk usage stuff?

Yes, exactly! If you want to do this as flatpaks, no problem, I'll close this PR and make it work with flatpaks.

castrojo:
Yeah on Bluefin we already moved to loupe from eog and the text editor is in there and the rest of the coreapps are mostly the same as Silverblue.

ArtikusHG:
So you think we should make a yafti config for main for this? If so, I'll probably close this PR and make a new one

castrojo:
yafti isn't in main, those images are "light touch": https://universal-blue.org/scope/#main-images

ArtikusHG:
Yeah I got a bit confused, sorry. How are flatpak installations handled in main? It doesn't seem to have a /etc/flatpaks/user/install file either, I must be missing something

castrojo:
They're not, we don't touch the flatpak setup in the main images, I haven't checked in a while but I think you get the fedora flatpak remote by default?

castrojo:
Actually I think in stock Fedora they're installed as part of the installation process before first boot.

ArtikusHG:
So how should we add a selection of default flatpak apps to main? Do we have do it in the installer?

castrojo:
Yeah but we're in the middle of figuring out how to do that, but I don't think we'd make offline ISOs for all the main images anyway. They're all larger than what we can host on github. Once the OCI support in the installer is better upstream then we could make an online installer that we could host here that would replicate the functionality but we haven't been able to figure that out yet (multiple people are trying to figure it out). 

I guess you could port the service units in bazzite/bluefin to install the packages in the background on first boot but that'd also be problematic as we'd have to add a bunch of stuff to main that people would probably not like (installing packages in the background without prompting, etc.).

ArtikusHG:
I think being able to install flatpaks at first boot would be great, both for the images themselves and for the people who then fork them and make custom images. If you are concerned that people won't like it, perhaps we could do a vote?

Anyhow, I feel like this issue really needs to be solved somehow, as main images are way too barebones and not really usable out of the box without a few manual app installations (at least that's the case for Silverblue). If this will be solved in the iso/installer in the future, I can close this, but I feel like porting the first boot flatpak installer is the way to go.

bsherman:
>  If this will be solved in the iso/installer in the future, I can close this, but I feel like porting the first boot flatpak installer is the way to go.

I think this PR needs to be closed regardless of future plans, and instead make it an issue. 


Related to the topic, my personal position is the current ISO installer is like pre-alpha, while our main images themselves are beta... and the best way to use Universal Blue is to install silverblue/kinoite/etc from official ISO, then rebase.  This somewhat solves the concern.

When we have reliable ISO/installer solutions, that's when I'm concerned about ensuring the first boot experience of `*-main` installs matches upstream. But, I'd want it to do just that, do what upstream does, not re-invent that wheel. Getting fun with Yafti and things is all for Bluefin, Bazzite, and other downstreams people want to create.

ArtikusHG:
Alright, then I'm closing this and hoping to get it solved in the installer.

bsherman:
> Alright, then I'm closing this and hoping to get it solved in the installer.

Hope it's clear, I do support creating an issue to track the concern.  

If you do, please link it here for context. Thanks!

PR #383, opened 2023-10-08
packages: Add Gstreamer plugins

EliasOfWaffle:


bigpod98 (commented):
Is this already part of upstream packages in worstation?

EliasOfWaffle:
> Is this already part of upstream packages in fedoea workstation?

Gstreamer is already pre installed i think but the vaapi and bad plugins not, but it can be part of the scope because is Non Free codec enablement
"
Non-free codecs aka. "a bunch of that RPMFusion stuff".
"

PR #384, opened 2023-10-08
post-install: Enable Fstrim.timer systemctl service

EliasOfWaffle:


KyleGospo:
This should already be enabled upstream.

EliasOfWaffle:
> This should already be enabled upstream.

for some reason i have to enabled in my pc, can be necessary to view if is really enabled as default using "sudo systemctl status fstrim.timer"

KyleGospo:
https://fedoraproject.org/wiki/Changes/EnableFSTrimTimer

Enabled in Fedora 32

EliasOfWaffle:
> https://fedoraproject.org/wiki/Changes/EnableFSTrimTimer
> 
> Enabled in Fedora 32, also enabled in a new VM.

oh thanks, in this case my fedora 39 beta can be a bug ? i think that a can be make a confusion about, thanks.
In this case i will close this PR, sorry for my mistake

PR #385, opened 2023-10-08
packages: Add Mesa3D OpenCL driver

EliasOfWaffle:


bigpod98 (commented):
Is this already part of upstream packages in worstation?

EliasOfWaffle:
> Is this already part of upstream packages in worstation?

is part of fedora repos, but is not installed as default in nothing ISO, i think that OpenCL looks much optional for be a default preinstalled in fedora, but the Mesa-LibOpenCL already have Mesa3D RustICL Gallium StateTracker that is conformance with OpenCL 3.0 in Intel Iris Driver, and in future it will be support Zink driver

bigpod98:
still if it isnt installed with any iso specially fedora Workstation i dont think has my approval at this time based on scope found at https://universal-blue.org/scope/

EliasOfWaffle:
> still if it isnt installed with any iso specially fedora Workstation i dont think this should have my approval at this time based on scope found at https://universal-blue.org/scope/

it can be considered a hardware enablement? because OpenCL is already used in various softwares including Darktable and OpenCL is a heterogeneous API for GPU enablement, in this case Iris driver can use OpenGL API that is a thing that Intel GPUs support.
In case if this can be considered the Intel-gmmlib, intel computer runtime zero level for Intel OneAPI support for intel GPU can be considered a hardware enablement?

bigpod98:
While yes this can be cinsidered hardwrae enablement could this be packckaged as a part of flatpak or oci image with the peogram using it if so this likely doeant apply as those are recommended methods of installation of software plus installing it here would likely have no effect on programms running inside flatpak and podman(or other oci)

KyleGospo:
-1, this works in distrobox and makes alternatives like ROCM harder to use

Issue #394, opened 2023-10-17
Add nautilus-python & nemo-python packages


fiftydinar:
### Describe the package

Needed for Nautilus & Nemo file manager Python extension integration.

Also needed for popular GSConnect extension, for "Send to mobile device" context menu.

### Information on the package

Package filename: nautilus-python-4.0-2.fc38.x86_64.rpm
Package name: nautilus-python
Package version: 4.0
Package release: 2.fc38
Homepage: https://wiki.gnome.org/Projects/NautilusPython
License: GPLv2+
Download size: 30.26 KB
Installed size: 62.56 KB

Package filename: nemo-python-5.6.0-2.fc38.x86_64.rpm
Package name: nemo-python
Package version: 5.6.0
Package release: 2.fc38

### Image

Silverblue & Cinnamon

fiftydinar:
Pull request is closed, so close this too.

PR #395, opened 2023-10-17
feat: Add nautilus-python

fiftydinar:
Needed for Nautilus file manager Python extension integration.

Also needed for popular GSConnect extension, for "Send to mobile device" context menu.

Issue: https://github.com/ublue-os/main/issues/394

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -46,7 +46,8 @@
             ],
             "silverblue": [
                 "adw-gtk3-theme",
-                "gnome-tweaks"
+                "gnome-tweaks",
+                "nautilus-python"
             ],
             "kinoite": [
                 "icoutils",


bsherman:
I'm not sure this fits our scope for inclusion in `*-main` images: https://universal-blue.org/scope/#main-images

ArtikusHG:
> not sure this fits our scope for inclusion in *-main images
imo it does, since it is required for GSConnect and probably a bunch of other convenience things. people expect this kind of stuff to be on the image, and it's not like this package is a huge one

castrojo:
I'm -1 on this one as well. 

These are base images and we purposely scope them down so to avoid having to include all the packages that people think they need. 

The right place for this to be seeded is either in upstream Fedora or in a downstream image.

PR #401, opened 2023-10-27
fix: stop excluding kwrite

### LINKED ISSUE #400: Revert kate integration over kwrite in Kinoite
### Describe the package

The inclusion of `kate` over `kwrite` is an opinionated decision that diverges from upstream. KWrite serves as a simpler text editor without extra features that are targeted only to developers.

#257 argues that Kate "would be more useful for ublue-os demographics", but I don't see how the target demographics of ublue main images should be different from Kinoite's. Therefore, the decision of including one package over another should be, in any case, from Kinoite (or a downstream fork).

### Information on the package

```
Name         : kwrite
Version      : 23.08.1
Release      : 1.fc38
Architecture : x86_64
Size         : 253 k
Source       : kate-23.08.1-1.fc38.src.rpm
Repository   : @System
From repo    : updates
Summary      : Text Editor
URL          : https://apps.kde.org/kate/
License      : LGPLv2+
Description  : Text Editor.
```

### Image

Kinoite

castrojo:
Fixes #400

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -214,8 +214,7 @@
             ],
             "kinoite": [
                 "plasma-discover-rpm-ostree",
-                "ffmpegthumbnailer",
-                "kwrite"
+                "ffmpegthumbnailer"
             ],
             "mate": [
                 "python3-unbound",


bsherman (approved):
I agree that kwrite is part of upstream and shouldn't be removed on this main image.

Issue #363, opened 2023-09-24
Rebase silverblue, kinoite, onyx, and sericea to official images


castrojo:
Lots of the stuff got punted but at least we have official quay.io/fedora images, so let's selectively switch what we can to do this. I'm assuming onyx will be up there for F39.

Huge milestone folks, this starts to line us up with images tested by Fedora QA. 
 
This is a partial fix for:

https://github.com/ublue-os/main/issues/78

@bsherman wanted to work on this so you get first dibs I guess!

IverCoder:
So we will be working closer to upstream now? Can't wait for this to be finished ❤️

castrojo:
Just to set expectations no one's really working on this, this would have to come from Fedora.

castrojo:
Oh sorry, wrong one, I meant to leave this comment on the download efficiency one, someone is going to work on this. 😄

castrojo:
This is the issue to track: https://pagure.io/pungi/pull-request/1699

bsherman:
Note that onyx is not yet being published  here: https://quay.io/organization/fedora

So the PR for this covers Silverblue, Kinoite, and Sericea only.

https://github.com/ublue-os/main/pull/407

bsherman:
@castrojo @EyeCantCU and I worked up a cron schedule based on the timings of official upstream images:

```
# cron adjustments

1330 config
1400 akmods (built on silverblue)
1500 main
1530 nvidia
1530 framework
1540 surface, asus
1610 surface-nvidia, asus-nvidia
1645 bazzite/bluefin/beyond
```

PR #407, opened 2023-10-31
feat(f39): Use official images for Silverblue, Kinoite, and Sericea

EyeCantCU:
Uses the official images provided by Fedora for Silverblue, Kinoite, and Sericea found here:

https://quay.io/organization/fedora

This excludes other images as well as Fedora 38 and prior as official images are not available.

Closes: https://github.com/ublue-os/main/issues/363


bsherman:
@EyeCantCU I made a few small fixes, should be covered in my commit message and in my update to the linked issue.

castrojo:
What's the tldr on the cron adjustments needed?

bsherman:
> What's the tldr on the cron adjustments needed?
My ballpark guess is we need to push back the cron for everything by 10-13 hours. 

But I had to step away and this needs to be verified.

castrojo (approved):

PR #410, opened 2023-11-02
fix: Target Fedora 40 for official images, revert 39 to ci-test

EyeCantCU:
Official images aren't quite ready yet


bsherman (approved):
I like this solution, and we can bump back down to 39 if Fedora tells us those images are good to go.

Issue #187, opened 2023-04-27
Boots into blank screen on tty1


Doosty:
This image boots into black screen, switching to any other tty works. Bsherman on discord gave me a solution to enable this systemd service either at runtime or inside Containerfile `RUN systemctl enable getty@tty1`.

castrojo:
Note: this is for base-main, which doesn't come with a desktop. We should document this.

manan-gup:
I ran into this yesterday and randomly tried switching to a different tty and it worked (like its mentioned above). But I was almost about to reinstall with another version. Its not obvious whether the system is bricked or not. 

If its as simple as adding this to the Containerfile then we can change it instead of documenting a workaround, right?
> `RUN systemctl enable getty@tty1`

castrojo:
Yeah I think if someone PRs in for this just for -base that would do the trick?

manan-gup:
I'll do it. Thanks for the quick response. Is reading the stuff [here](https://universal-blue.discourse.group/c/contributing/7) enough to set me up to begin contributing?

castrojo:
https://universal-blue.org/CONTRIBUTING/ is the Canonical text, though I should have pointers to that from the forum, I'll make sure I get on that, thanks!

PR #424, opened 2023-11-15
fix: enable boot to tty for base image

manan-gup:
Added statement to enable automatically booting to a tty for the base image without a DE as per #187.

I hope I've added the statement at the right place. Also, I wasn't sure what the idiomatic way to check for the base image was so I used an if statement. Let me know if I need to change anything. Thanks.


EyeCantCU (approved):
This looks fantastic and is the proper way to go about it. Thank you so much for the contribution!

manan-gup:
Thanks to everyone for the kind and generous experience. This was my first real, though small, open source contribution. :)

PR #436, opened 2023-11-30
fix(ci): never skip successful builds check

p5:
Thanks @bobslept for this fix!  Copying from Bluefin to main so the check step always runs - even if a build fails.

Currently a build may fail and as this check step is skipped, we can merge without this check job being successful.

https://github.com/ublue-os/bluefin/pull/696


bobslept (approved):
LGTM :+1:

bsherman (approved):

PR #444, opened 2023-12-07
chore(podman): Remove the workaround for podman

john-ghatas:
Seeing the package page of Fedora podman 4.8.1 has dropped in the stable branch; https://src.fedoraproject.org/rpms/podman. 

I propose we remove this workaround whenever is viable

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -19,11 +19,6 @@ COPY github-release-install.sh \
 COPY --from=ghcr.io/ublue-os/config:latest /rpms /tmp/rpms
 COPY --from=ghcr.io/ublue-os/akmods:main-${FEDORA_MAJOR_VERSION} /rpms/ublue-os /tmp/rpms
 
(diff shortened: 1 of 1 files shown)


castrojo:
Thanks! 

(For the approver): Let's just make sure we verify that the package is in the oci image we derive from so we don't remove it too early!

bobslept (approved):
Seems to be safe to remove the workaround.

```
user@fedora:~$ rpm-ostree status
State: idle
Deployments:
● ostree-unverified-registry:quay.io/fedora-ostree-desktops/silverblue:39
                   Digest: sha256:f66a5cfcf1fed1e155c8e8a4a83b408e13d8a545f966689b03b5e1ea6ed36114
                  Version: 39.20231209.0 (2023-12-09T02:11:50Z)
user@fedora:~$ rpm -q podman
podman-4.8.1-1.fc39.x86_64
user@fedora:~$ podman -v
podman version 4.8.1
user@fedora:~$ 
```

castrojo (approved):

PR #474, opened 2024-01-16
fix: Enforce 64-bit version of mesa-va-drivers-freeworld

KyleGospo:
Potentially solves an issue where, when rpmfusion is out of date, this will install the 32-bit version of the package and not fail as it should.

As a result, this PR should fail to build until the package is built upstream. We'd rather not update than update w/ broken builds.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -24,7 +24,7 @@
                 "libva-utils",
                 "lshw",
                 "mesa-filesystem",
-                "mesa-va-drivers-freeworld",
+                "mesa-va-drivers-freeworld.x86_64",
                 "net-tools",
                 "pam_yubico",
                 "pam-u2f",
--- a/packages.json
+++ b/packages.json
@@ -23,7 +23,6 @@
                 "libva-intel-driver",
                 "libva-utils",
                 "lshw",
-                "mesa-filesystem",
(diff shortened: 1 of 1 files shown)


KyleGospo:
Confirmed this fails to build, marking as ready for review as a result. This won't merge unless everything build successfully which is the intended behavior

bsherman (approved):
This solution makes sense. I agree we'll have to get a little smarter about it when we eventually add `aarch64`.

bsherman:
FYI for @ublue-os/approver : please do not force merge this PR. It should only be merged once a full set of clean build checks can run.

bsherman (approved):

castrojo (approved):

bsherman:
> FYI for @ublue-os/approver : please do not force merge this PR. It should only be merged once a full set of clean build checks can run.

And... in Discord Jorge, Kyle, Noel and I discussed further.  We wanted to avoid breaking builds with this PR, but, all builds are now broken due to upstream Fedora/RPMfusion package updates. So this seems a good time to merge as we won't break things worse, but should get a clean build once the dust settles. So this can now be force merged.

KyleGospo:
BAH GOD IT'S KYLE WITH A STEEL FORCE MERGE

PR #481, opened 2024-01-22
feat: Enable multi-arch builds

ponzienjoyer:
This change enables multi-arch builds with support for amd64 and aarch64

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,6 +31,7 @@ jobs:
           - vauxite
         major_version: [38, 39]
         build_target: [nokmods, kmods]
+        arch: [amd64, aarch64]
         include:
           - major_version: 38
             is_latest_version: false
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -92,14 +93,17 @@ jobs:
...
+          echo "BUILD_ARCH=${{ matrix.arch }}" >> $GITHUB_ENV
 
       - name: Generate tags
         id: generate-tags
         shell: bash
         run: |
           # Generate a timestamp for creating an image version history
           TIMESTAMP="$(date +%Y%m%d)"
-          VARIANT="${{ matrix.major_version }}"
+          VARIANT="${{ matrix.major_version }}"-"${{ matrix.arch }}"
```
ponzienjoyer, line 104: i wasn't sure if this was useful or if splitting the values was better. Opened the PR like this in case folks had a preference

ponzienjoyer:
Have some pretty rough edges with my build yml changes, so I'm closing the PR for now to fix.

Issue #470, opened 2024-01-10
Override default systemd emergency service behavior with coreos-su-login-force.conf


noelmiller:
> > In the short term I'd definitely advocate for Fedora and derivatives to carry that change by default and tell anyone making kisoks etc. that they need to disable it. We just chose to carry it in FCOS derivatives by default currently.
> 
> We may want to consider shipping this by default for our images. Thoughts? @bsherman @castrojo @KyleGospo

I support adding the `coreos-sulogin-force.conf` from  https://github.com/coreos/fedora-coreos-config/blob/17b7f15f49c00ba1e522f15a72f6df70db01fdcc/overlay.d/05core/usr/lib/systemd/system/emergency.service.d/coreos-sulogin-force.conf#L1 to our desktop builds, from main on up.

This avoids current hassle for users which doesn't provide any real security benefit.

_Originally posted by @bsherman in https://github.com/ublue-os/main/issues/469#issuecomment-1885875698_

bsherman:
@travier you asked for someone to reach out in [your comment](https://github.com/ublue-os/main/issues/469#issuecomment-1887184104). 

I started poking around https://github.com/coreos/fedora-coreos-config/blob/stable/overlay.d/05core/usr/lib/systemd/system/emergency.service.d/coreos-sulogin-force.conf to see how this was being done on CoreOS. But I'm unsure how to get started adding this into the atomic desktops.

We can connect however you prefer, here, discord, email. My email is on my Fedora people profile (`@bsherman1` not the same as my github handle).

Thank you for the offer to assist!

travier:
So I've dived into this more and I've found https://github.com/coreos/fedora-coreos-tracker/issues/805#issuecomment-1888794647, which makes it a harder sell for switching it on for all Fedora desktops.

I would probably be a good idea to start with a discussion thread on discourse or on fedora-devel to get more ideas on how this could be safely done.

travier:
In general, the process to get such a changes in Fedora is to make a Change Request to have it be visible to the community and force the discussion to happen (and a decision to be taken).

Writing / drafting a change page following the instructions in https://docs.fedoraproject.org/en-US/program_management/changes_policy/ is good first step.

bsherman:
> So I've dived into this more and I've found [coreos/fedora-coreos-tracker#805 (comment)](https://github.com/coreos/fedora-coreos-tracker/issues/805#issuecomment-1888794647), which makes it a harder sell for switching it on for all Fedora desktops.

[...]

PR #488, opened 2024-02-04
feat: allow rescue/emergency boot with grub cmdline args

bsherman:
This uses a systemd-generator to dyamically write a drop-in config for the rescue and emergency services only when they are requested via the kernel cmdline, which requires console/grub access. This allows use of these modes with the default Fedora state of a password locked root user, but does not auto-allow root access in the case of a failed fsck-check, which can also drop into the emergency shell.

Relates: #470

### DIFF
--- a/sys_files/usr/lib/systemd/system-generators/ublue-sulogin-generator
+++ b/sys_files/usr/lib/systemd/system-generators/ublue-sulogin-generator
@@ -1,0 +1,65 @@
+#!/usr/bin/bash
+
+# This systemd.generator(7) detects if rescue or emergency targets were
+# requested from the kernel cmdline; if so, it overrides the respective
+# target to set force sulogin, allowing use of rescue/emergency targets
+# on systems with locked root password (as is Fedora default).
+#
+# This does NOT bypass locked root password on a fsck failure, but WILL
+# bypass when rescue/emergency targets are chosen from kernel cmdline.
+# Since this requires console/grub access, it is assumed to be at least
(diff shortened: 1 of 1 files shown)


bsherman:
@ublue-os/approver please do NOT merge this until we've collected feedback. :-) Thank you.

cgwalters:
Looks sane to me.  Arguably we could try to put this in systemd upstream too.

bsherman:
> Looks sane to me. Arguably we could try to put this in systemd upstream too.

I'm not sure how easy it is to get merged into different projects. At the least, I figured we'd include it here and I'll PR to CoreOS.

Then we'd have running use cases before pushing further upstream.

KyleGospo (approved):
Tested well on my end!

TriMoon:
Not to be nit-picking but....
Title should be "kernel" instead of "grub", as it is not specific to Grub...

TriMoon (commented):
Not to be nit-picking but....
Title should be "kernel" instead of "grub", as it is not specific to Grub...

Plus it will boot, i guess, in a root shell?
_(So mention that also)_

EyeCantCU (approved):
Works over here as well :)

bsherman:
> Arguably we could try to put this in systemd upstream too.

@cgwalters , I had planned to PR this into CoreOS as a replacement for the current static drop-ins. Would prefer I first attempt going upstream to systemd?

cgwalters:
Landing in FCOS first sounds totally sane to me.

PR #495, opened 2024-02-09
chore(ci): conditional pin to repo variable RPMFUSION_MIRROR

bsherman:
Prompted by ocf.berkeley being down for a system rebuildand the proven usefulness of using a single mirror, this approach allows local builds with no build-args to work with a default RPMFusion mirror metalink setup, but in case of the build arg (in our CI, a repo variable) RPMFUSION_MIRROR being set, that URL will be substituted as the forced single mirror repo.


p5 (approved):
Nice one!

PR #497, opened 2024-02-09
chore: enable rpmfusion testing repos

bsherman:
Enabling rpmfusion-*-updates-testing repos with a lower priority (110 vs the default of 99, as lower is higher priority with yum/dnf/rpm-ostree) such that the testing packages (new mesa version) can get picked up sooner by our builds, but the repo should not be preferred for upgrades unless the version in the repo is specifically required.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -12,6 +12,12 @@ rpm-ostree install \
     /tmp/rpms/*.rpm \
     fedora-repos-archive
 
+if [[ "${FEDORA_MAJOR_VERSION}" -ge 39 ]]; then
+    # note: this is done before single mirror hack to ensure this persists in image and is not reset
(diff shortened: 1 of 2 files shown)


p5 (approved):

p5:
Approved, but this sort of thing may need more than one approval

bketelsen (approved):
I think this is fine. our builds will catch the minimal amount of trouble this could bring.

KyleGospo (approved):

bsherman:
> I think this is fine. our builds will catch the minimal amount of trouble this could bring.

Yep, I agree... especially with a lower priority on rpmfusion testing... I don't see a huge amount of trouble.

Also, for context, small number of packages here: https://mirror.fcix.net/rpmfusion/free/fedora/updates/testing/39/x86_64/repoview/index.html

PR #499, opened 2024-02-12
build: decrease image sizes by utilizing buildah commands

### LINKED ISSUE #498: Use better defined buildah techniques
Noel's friend was working on this. We're just using the vanilla buildah action, ends up you can optimize. This is at least 600MB, we could probably squeeze out some savings here:

https://github.com/JasonN3/ublue-os-main/pull/1

```
REPOSITORY                                                   TAG         IMAGE ID      CREATED         SIZE
ghcr.io/jasonn3/base                                         6854aa4-38  3bd3c54caee4  15 minutes ago  4.04 GB
ghcr.io/ublue-os/base-main                                   latest      a63c9e5d4a59  3 hours ago     4.69 GB
```

JasonN3:
The existing build process just uses `buildah build`. This command is similar to `docker build` and `podman build` and will build a new layer for each command in the file, which means files can never be removed from the image. They can only be hidden from future layers. Using a buildah script that utilizes the rest of the buildah commands, you can build a single image layer with multiple commands. This gives you the ability to add files to the image, run commands, and then remove the files from the image before it committed as a layer.

Noel mentioned that bazzite is another image that ends up pretty large, so I'll do the same conversion for that Containerfile and submit a PR for that repo as well.

Fixes #498


Containerfile.sh
p5, line 1: Question: How does this differ from using something like `--squash` Buildah option?

JasonN3: I hadn't come across `--squash`. It does look like it would work similarly. However, one additional benefit you get from the multiple commands approach is that you can run things outside of the container. For example, in bazzite, you are doing some git clones and downloading some tar.gz files. These actions could be done outside of the image so you wouldn't need to install the related packages.

JasonN3: Actually, I just looked at the current docs. Apparently even though there was an issue were buildah agreed that `--squash` shouldn't squash the base layers, it still does. This means that when pulling the image, it would appear as a single giant layer and you would be able to cache the base image separately.

p5:
Happy with this change, though I'm going to wait for others to look at it since we're changing from a Dockerfile (which everyone knows how it works) to something a bit more complex.  
 
I don't know enough about how Buildah works in a scripted way to give my opinion.

JasonN3:
Converting this to a draft because I found 2 issues with it.
1. `$(rpm -E %fedora)` gets run outside of the container, which results in it not resolving. This is easy to fix since there is a variable for the version.
1. All of the `&&` cause commands to run outside of the container. I should have replaced all of those but I tried to do a simple replace.

JasonN3:
> To clarify, the main advantage of using buildah in this way is it will not create layers until you have run the `buildah push` command at the end of the script? It should be one giant single layer on the base image?

When you run `buildah commit`, but otherwise yes. However, the existing workflow apparently also only commits a single layer. See below.

> Secondly, since commands are run outside the container, you do not need to include dependencies in the image since those commands are run outside it?

Correct. If there are any packages needed only for the building process, they would not need to be included in the resulting image.


I fixed up the issues I found and that surfaced a few other issues. Apparently none of the packages were installing initially even though the workflow step was passing. I fixed that up and once the packages installed, the resulting image was 4.69GB, which is the same size as the image on this repo. This points to that Buildah is squashing the Containerfile into one layer already. I updated the fork to include the fixed version. 
Considering there is no size benefit, it is up to you guys on whether you would like the added benefit of being able to run commands outside of the container while building it. If not, feel free to close this PR.

bsherman:
@JasonN3, I very much appreciate the effort in trying to help find some space savings, but as they didn't materialize... I'm voting "no" in the change.

I believe it is desirable to use Containerfile as it is a well-known pattern for docker/podman users. So moving to a scripted podman approach would need to have compelling benefit.

One thing I have learned from the process of evaluating your efforts... It is possible we could still use `Containerfile` but forgo the use of `redhat-actions/buildah-build` if it lets us do things like `podman build --squash -f Containerfile` .  

And... either the approach I mentioned here, or the overall `scripting podman` approach, could prove more useful for Bazzite or Bluefin images than it did here for main.

bsherman:
I'm closing this PR per discussions in Discord and the fact this effort didn't result in a size benefit.

PR #500, opened 2024-02-12
fix: add pcscd polkit rule for Yubikey access

rothgar:
I'm not exactly sure if this should be in /etc/polkit-1/rules.d or /usr/share/polkit-1/rules.d

I put it in /etc/ because it affects a user and this would allow them to change/delete the rule if needed.


bsherman:
@rothgar thanks for the PR!

I wouldn't add this here, instead, we've been adding rules like this to our config RPMs so they can be tracked via RPM database.

Side note: the only directly added file in this repo is our systemd-generator for sulogin. Maybe I shouldn't have even added it here, but I did so as I deem it temporary due to our goal of getting it added upstream to either Fedora or systemd itself.


It looks like your solution could be a fix to this issue there: https://github.com/ublue-os/config/issues/180

I want to close this PR, but would you mind submitting a PR to https://github.com/ublue-os/config ?

It's a bit more involved as it will include tweaking the RPM spec, but I'm happy to walk you through any questions on a PR there.

Issue #502, opened 2024-02-18
Automated retries?


dylanmtaylor:
I see that build actions sometimes fail. I think we should leverage the retry action on ublue builds with an attempt limit of 3.
 https://github.com/marketplace/actions/retry-action

That way if it's a weird network issue it something we won't have a day without a new image.

bsherman:
I've added requested changes to the associated PR ( #503 ), but I'll add some thoughts here for good measure.

@dylanmtaylor is very much correct that we sometimes have spurious failures (various causes which likely include network issues), and those result in the team needing to manually retry some runs of the workflow. As an example, a spurious failure will usually result in success for most of the matrix options, but one or two will fail. 

I do agree that automatically retrying certain steps of the workflow will be helpful.

I identified the two most useful in my requested changes:
1.  **Get current version** was recently discovered to have intermittent but **SILENT** failures, so I fixed it to actually halt workflow execution rather than result in images which contain incomplete metadata. This is a simple step and a great candidate for automatic retry. In every failure case I've seen, it has been clear that it was some intermittent issue in network or on the ghcr servers which caused the failure.
2.  **Push to GHCR** is also a great candidate, since, if we reach this step, all that must be done is publish the cleanly built image. However, this does fail on occasion, and the fault is always some issue with network or on the GHCR server side. Likely our very large image sizes don't help us here.

I've specifically requested that we do NOT auto-retry the most complex step, **Build Image**. The most common causes of failure here are legitimate, usually due to an upstream RPM dependency issue.  The one spurious issue I do know of in **Build Image** is related to the `github-release-install.sh` shell script which helps us install RPM packages direct from a project's github release. This is where I'd like to see an improvement to the shell script to handle those failures and retry internally. I've already made one such attempt with only partial success.


In addition to all this, I'd really like to see these improvements in `ublue-os/main`... but I hesitate to implement in the 6 other "foundational"/"hardware enablement" repos we maintain. We've already had some discussions on merging and cleaning them up as it's currently very messy to maintain them all as distinct repos.

Hope that provides some context to any reader regarding my views on this topic.

bsherman:
Actually, i think we should close this as "done" since we merged the PR at the top and have continued to add appropriate retry logic in various places throughout the project.

PR #503, opened 2024-02-18
feat: leverage retry-action to increase reliability of builds.

dylanmtaylor:
This is a proposed implementation of #502. It needs to run the GitHub action to be tested, but should improve reliability without manual build re-triggers.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -240,7 +240,7 @@ jobs:
         run: |
           cosign sign -y --key env://COSIGN_PRIVATE_KEY ${{ steps.registry_case.outputs.lowercase }}/${{ steps.build_image.outputs.image }}@${TAGS}
         env:
-          TAGS: ${{ fromJSON(steps.push.outputs.outputs).digest }}
+          TAGS: ${{ steps.push.outputs.outputs && fromJSON(steps.push.outputs.outputs).digest }}
           COSIGN_EXPERIMENTAL: false
           COSIGN_PRIVATE_KEY: ${{ secrets.SIGNING_SECRET }}
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -147,15 +147,18 @@ jobs:
 
       - name: Get current version
         id: labels
-        shell: bash
-        run: |
-          set -eo pipefail
-          ver=$(skopeo inspect docker://quay.io/${{ env.SOURCE_ORG }}/${{ env.SOURCE_IMAGE }}:${{ matrix.major_version }} | jq -r '.Labels["org.opencontainers.image.version"]')
-          if [ -z "$ver" ] || [ "null" = "$ver" ]; then
-            echo "inspected image version must not be empty or null"
-            exit 1
-          fi
-          echo "VERSION=$ver" >> $GITHUB_OUTPUT
+        uses: Wandalen/wretry.action@v1.4.4
+        with:
+          attempt_limit: 3
+          attempt_delay: 15000
+          command: |
+            set -eo pipefail
+            ver=$(skopeo inspect docker://quay.io/${{ env.SOURCE_ORG }}/${{ env.SOURCE_IMAGE }}:${{ matrix.major_version }} | jq -r '.Labels["org.opencontainers.image.version"]')
+            if [ -z "$ver" ] || [ "null" = "$ver" ]; then
+              echo "inspected image version must not be empty or null"
+              exit 1
+            fi
+            echo "SOURCE_IMAGE_VERSION=$ver" >> $GITHUB_ENV
 
       # Generate image metadata
       - name: Image Metadata
@@ -166,7 +169,7 @@ jobs:
             ${{ env.IMAGE_NAME }}
           labels: |
             org.opencontainers.image.title=${{ env.IMAGE_NAME }}
-            org.opencontainers.image.version=${{ steps.labels.outputs.VERSION }}
+            org.opencontainers.image.version=${{ env.SOURCE_IMAGE_VERSION }}
             org.opencontainers.image.description=A base Universal Blue ${{ matrix.image_name }} image with batteries included
             io.artifacthub.package.readme-url=https://raw.githubusercontent.com/${{ github.repository }}/main/README.md
             io.artifacthub.package.logo-url=https://avatars.githubusercontent.com/u/120078124?s=200&v=4
@@ -201,20 +204,24 @@ jobs:
           string: ${{ env.IMAGE_REGISTRY }}
 
       - name: Push To GHCR
-        uses: redhat-actions/push-to-registry@v2
+        uses: Wandalen/wretry.action@v1.4.4
         id: push
         if: github.event_name != 'pull_request'
         env:
           REGISTRY_USER: ${{ github.actor }}
           REGISTRY_PASSWORD: ${{ github.token }}
         with:
-          image: ${{ steps.build_image.outputs.image }}
-          tags: ${{ steps.build_image.outputs.tags }}
-          registry: ${{ steps.registry_case.outputs.lowercase }}
-          username: ${{ env.REGISTRY_USER }}
-          password: ${{ env.REGISTRY_PASSWORD }}
-          extra-args: |
-            --disable-content-trust
+          action: redhat-actions/push-to-registry@v2
+          attempt_limit: 3
+          attempt_delay: 15000
+          with: |
+            image: ${{ steps.build_image.outputs.image }}
+            tags: ${{ steps.build_image.outputs.tags }}
+            registry: ${{ steps.registry_case.outputs.lowercase }}
+            username: ${{ env.REGISTRY_USER }}
+            password: ${{ env.REGISTRY_PASSWORD }}
+            extra-args: |
+              --disable-content-trust
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v3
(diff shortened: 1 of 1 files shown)


dylanmtaylor:
I need to figure out why this is failing before it's merged.

dylanmtaylor:
The issue is related to this section

            labels: |
              ${{ steps.meta.outputs.labels }}
              
              I think the retry action is sensitive to the indentation here.

dylanmtaylor:
This is now fixed using a workaround:

I pad labels in another step and output it to the environment. This is necessary because we need it to remain a multiline string without using any character substitution. I then pull the string into the inputs for the retry-action. You can confirm the inputs work on the build step.

dylanmtaylor:
It's not the cleanest in the world, but if it gets us working retries, that is a huge win.

```
      # Workaround to fix indentation for the labels so that retries work.
      - name: Left pad labels
        id: left_pad_labels
        shell: bash
        env:
          LABELS: ${{steps.meta.outputs.labels}}
        run: |
          echo "Labels:"
          echo "${LABELS}"
          export TRANSFORMED=$(echo "$LABELS" | sed 's,^,  ,g')
          echo "Transformed:"
          echo "${TRANSFORMED}"
          echo "LABELS<<EOF" >> $GITHUB_ENV
          echo "$TRANSFORMED" >> $GITHUB_ENV
          echo "EOF" >> $GITHUB_ENV
```

.github/workflows/build.yml
```diff
@@ -171,27 +171,50 @@ jobs:
             io.artifacthub.package.readme-url=https://raw.githubusercontent.com/${{ github.repository }}/main/README.md
             io.artifacthub.package.logo-url=https://avatars.githubusercontent.com/u/120078124?s=200&v=4
 
+      # Workaround to fix indentation for the labels so that retries work.
```
bsherman, line 174: Can you help me understand why this left padding is required?

bsherman: I see you have comments about the white space issue already.

dylanmtaylor: Without it, it evalutes the YAML to 

```
with:
    action: redhat-actions/buildah-build@v2
    with: containerfiles: |
    ./Containerfile
  image: kinoite-main
  tags: |
    pr-503-39 812996e-39 pr-503 812996e
  build-args: |
    IMAGE_NAME=kinoite
    SOURCE_ORG=fedora-ostree-desktops
    SOURCE_IMAGE=kinoite
    FEDORA_MAJOR_VERSION=39
    RPMFUSION_MIRROR=
  labels: io.artifacthub.package.logo-url=https://avatars.githubusercontent.com/u/120078124?s=200&v=4
  io.artifacthub.package.readme-url=https://raw.githubusercontent.com/ublue-os/main/main/README.md
  org.opencontainers.image.created=2024-02-18T09:21:11.721Z
  org.opencontainers.image.description=A base Universal Blue kinoite image with batteries included
  org.opencontainers.image.licenses=Apache-2.0
  org.opencontainers.image.revision=812996efb45acbbebbb18ac9ec5[67](https://github.com/ublue-os/main/actions/runs/7947996260/job/21697505580#step:8:68)a3fe[79](https://github.com/ublue-os/main/actions/runs/7947996260/job/21697505580#step:8:80)289fa
  org.opencontainers.image.source=https://github.com/ublue-os/main
  org.opencontainers.image.title=kinoite-main
  org.opencontainers.image.url=https://github.com/ublue-os/main
  org.opencontainers.image.version=39.20240218.0
```

This is not valid YAML, and breaks without valid indentation. Sadly, YAML is very sensitive to indentation, unlike JSON. By adding two spaces to the front of each line in that env variable and then passing it in, it is properly indented.

bsherman:
I'm not confident we want to always retry the image builds. Usually when image builds fail it is due to legit issues with package dependencies (rpmfusion etc). There are a few places where scripts can fail in ways that warrant retry, but I'd like to see those particular bits improved instead. 

I DO think it would valuable to auto retry when the image push fails since those failures are almost always due to some random ghcr/github hiccup.

dylanmtaylor:
> I'm not confident we want to always retry the image builds. Usually when image builds fail it is due to legit issues with package dependencies (rpmfusion etc). There are a few places where scripts can fail in ways that warrant retry, but I'd like to see those particular bits improved instead.
> 
> I DO think it would valuable to auto retry when the image push fails since those failures are almost always due to some random ghcr/github hiccup.

This can be modified to make the push retry.

bsherman (changes requested):
Thank you for your desire to help improve the ublue build reliability! 

I do have some requests which I believe will make the solution more relevant to areas where I believe we most often see spurious failures.

List of requests follows:
1. when specifying `uses: Wandalen/wretry.action@master` please use an actual version selector not `master` (I believe `@1` would work). This will cause dependabot to alert us with PRs when a new major version update is available.
2. please add retry action to _Get current version_ `labels` step (this step is known to cause spurious failures)
3. please add retry action to _Push to GHCR_ `push` step (this step is known to cause spurious failures)
4. please remove retry action `build_image` step (I believe the few spurious failures I've seen in this step should be fixed in the respective shell scripts, not at the top level)
5. since not using retry action for `build_image` step, please remove the left padding step

dylanmtaylor:
I have force pushed a new version of this commit. @bsherman could you please review it?

.github/workflows/build.yml
```diff
@@ -147,15 +147,18 @@ jobs:
...
+          attempt_limit: 3
+          attempt_delay: 15000
+          command: |
+            set -eo pipefail
+            ver=$(skopeo inspect docker://quay.io/${{ env.SOURCE_ORG }}/${{ env.SOURCE_IMAGE }}:${{ matrix.major_version }} | jq -r '.Labels["org.opencontainers.image.version"]')
+            if [ -z "$ver" ] || [ "null" = "$ver" ]; then
+              echo "inspected image version must not be empty or null"
+              exit 1
+            fi
+            echo "VERSION=$ver" >> $GITHUB_ENV
```
bsherman, line 161: Since we already have versions in the environment and now this `VERSION` isn't scoped to the step, can we please rename to `SOURCE_IMAGE_VERSION`?

Thanks!

.github/workflows/build.yml
```diff
@@ -166,11 +169,11 @@ jobs:
             ${{ env.IMAGE_NAME }}
           labels: |
             org.opencontainers.image.title=${{ env.IMAGE_NAME }}
-            org.opencontainers.image.version=${{ steps.labels.outputs.VERSION }}
+            org.opencontainers.image.version=${{ env.VERSION }}
             org.opencontainers.image.description=A base Universal Blue ${{ matrix.image_name }} image with batteries included
             io.artifacthub.package.readme-url=https://raw.githubusercontent.com/${{ github.repository }}/main/README.md
             io.artifacthub.package.logo-url=https://avatars.githubusercontent.com/u/120078124?s=200&v=4
-
+
```
bsherman, line 176: I know it's a nitpick, but please remove these extra spaces and the extra new line at line 198 since they don't seem to contribute anything.

Thank you!

bsherman (changes requested):

dylanmtaylor:
Updated with the requested changes.

bsherman:
@dylanmtaylor why did you just force push?

dylanmtaylor:
> @dylanmtaylor why did you just force push?

I missed one line of whitespace change at the end of the file. I thought that it would be ideal to not have any changes to whitespace.

dylanmtaylor:
Commit difference: https://github.com/ublue-os/main/compare/2b8ffe22406934db5af86d8d6acd17b072414a29..5f8f7196899cc6ba2853cd236758a42b130bbe43

dylanmtaylor:
Interesting - the last attempt actually failed during the build action.

error creating build container: writing blob: storing blob to file "/var/tmp/storage1964624803/9": happened during read: unexpected EOF
time="2024-02-19T17:54:53Z" level=error msg="exit status 125"
Error: Error: buildah exited with code 125

dylanmtaylor:
I think it may be beneficial to add the retry to the build image after all. I am testing a revised version of this with it in there too. If it builds successfully the first time it'll do nothing, and if not it gets rerun only a couple of times.

dylanmtaylor:
I opened #506 as an alternative version of this which I feel is a bit better due to retrying the build action as well.

dylanmtaylor:
Converting to draft as I feel the merits of this vs #506 should be discussed before one is merged.

bsherman:
> > @dylanmtaylor why did you just force push?
> 
> I missed one line of whitespace change at the end of the file. I thought that it would be ideal to not have any changes to whitespace.

Understood.

In general, it's preferable to not force push in PRs as it removes context from code reviews, etc. We've been looking at using a tool to enforce policy across all ublue-os org repos, which would include preventing force-push, but for now, I'm just asking to please not do it anymore. :-)

Our auto-merge settings already squash commits, so it force push of pre-squashed commits doesn't really gain anything.

dylanmtaylor:
This is ready to go in, assuming it fully passes CI. @bsherman

dylanmtaylor:
I think I can see what happened here. 

Error: The template is not valid. .github/workflows/build.yml (Line: 243, Col: 17): Error reading JToken from JsonReader. Path '', line 0, position 0.

^ The step output is blank because we only run the push action on the main branch and not on PRs. Signatures should also not run on PRs due to this.

bsherman:
> Error: The template is not valid. .github/workflows/build.yml (Line: 243, Col: 17): Error reading JToken from JsonReader. Path '', line 0, position 0.

Hmm... annoying that the `fromJSON` function must be running in the template even if the step is skipped.

Oh well, thank you for fixing.

bsherman (approved):

EyeCantCU (approved):

PR #508, opened 2024-02-22
chore(ci): add image pull with retry

bsherman:
This is another in a series of PRs which have been proposed for #502 .

My thought is to identify areas in build_image which have spurious failures and fix them if reasonable within our scripting, or otherwise address issues, like this one.

The idea here is we fix the spurious failures pulling our (very large) base images, by pulling them to the build runner before using the buildah action.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -182,7 +182,7 @@ jobs:
           command: |
             # pull the base image used for FROM in containerfile so
             # we can retry on that unfortunately common failure case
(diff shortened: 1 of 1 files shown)


bsherman:
This still won't catch all issues, and it doesn't even try to pre-pull akmods or config image layers ... but we have other efforts in flight to address those concerns by moving to a dnf/yum repo instead.

castrojo (approved):

bsherman:
The failure on this currently is a kernel version incompatibility between a kmod and image kernel. Shouldn't block merging of this.

noelmiller (approved):

PR #521, opened 2024-03-17
feat: enable preview of Fedora 40 main builds

bsherman:
This PR enables F40 `*-main` images builds with minor caveats:

1. The F40 builds are `nokmods` images only, which is the norm since F39.
2. No `mate` images are built since our upstream isn't yet building F40.

**Important note:**
The naming conventions for F40 images (other than silverblue/kinoite) are confusing because of the Fedora naming change to "Atomic Desktop". I may want to discuss this before we start publishing images with wrong names.

Currently we have:
- sericea
- onyx
- lazurite
- mate
- vauxite

Fedora upstream is renaming `sericea` to `sway-atomic` and `onyx` to `budgie-atomic` (or something like that). So I think we should find out what their image names will really be and use that, but I'm open to suggestions.

Also, https://quay.io/organization/fedora-ostree-desktops is publishing Lazurite and Vauxite, so I guess we can keep going with those, but would we move them to the new naming convention we use with the official Sway and Budgie images?

And finally, we don't have a maintainer for Mate,, so I'm not sure what we want to do with it. I think we may drop it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,6 +55,9 @@ jobs:
             major_version: 39
           - image_name: lxqt
             major_version: 40
+          # There is currently no Fedora 40 version of mate
+          - image_name: mate
+            major_version: 40
           # THE FOLLOWING EXCLUDE IS MESSY BUT TEMPORARY UNTIL F38 IS GONE
           # see: https://github.com/ublue-os/main/issues/369
           # Fedora 39+ images do not include custom kmods (legacy)
@@ -70,7 +73,7 @@ jobs:
       - name: Matrix Variables
         shell: bash
         run: |
-          if [[ "${{ matrix.major_version }}" -ge "40" ]] && \
+          if [[ "${{ matrix.major_version }}" -ge "41" ]] && \
             grep "${{ matrix.image_name }}" <<< "silverblue, kinoite, sericea"; then
               echo "SOURCE_ORG=fedora" >> $GITHUB_ENV
               echo "SOURCE_IMAGE=fedora-${{ matrix.image_name }}" >> $GITHUB_ENV
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,7 +74,7 @@ jobs:
         shell: bash
         run: |
           if [[ "${{ matrix.major_version }}" -ge "41" ]] && \
-            grep "${{ matrix.image_name }}" <<< "silverblue, kinoite, sericea"; then
+            grep "${{ matrix.image_name }}" <<< "silverblue, kinoite, sericea, onyx"; then
               echo "SOURCE_ORG=fedora" >> $GITHUB_ENV
               echo "SOURCE_IMAGE=fedora-${{ matrix.image_name }}" >> $GITHUB_ENV
           else
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
           - lazurite
           - mate
           - vauxite
-        major_version: [38, 39]
+        major_version: [38, 39, 40]
         build_target: [nokmods, kmods]
         include:
           - major_version: 38
@@ -40,20 +40,28 @@ jobs:
             is_latest_version: true
             is_stable_version: true
             is_gts_version: false
+          - major_version: 40
+            is_latest_version: false
+            is_stable_version: false
+            is_gts_version: false
         exclude:
           # There is no Fedora 38 version of onyx or lazurite
           - image_name: onyx
             major_version: 38
           - image_name: lazurite
             major_version: 38
-          # There is no Fedora 39 version of lxqt as it was replaced by lazurite
+          # There is no Fedora 39+ version of lxqt as it was replaced by lazurite
           - image_name: lxqt
             major_version: 39
+          - image_name: lxqt
+            major_version: 40
           # THE FOLLOWING EXCLUDE IS MESSY BUT TEMPORARY UNTIL F38 IS GONE
           # see: https://github.com/ublue-os/main/issues/369
           # Fedora 39+ images do not include custom kmods (legacy)
           - build_target: kmods
             major_version: 39
+          - build_target: kmods
+            major_version: 40
     steps:
       # Checkout push-to-registry action GitHub repository
       - name: Checkout Push to Registry action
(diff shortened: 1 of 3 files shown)


bsherman:
Blocked by: https://github.com/ublue-os/akmods/pull/143

castrojo:
IMO let's name them what upstream Fedora names them, and then for things that timothee adds to the repo we follow we should just mirror as a community service.

Things that we don't have maintainer for like mate (and other repos in the org) we should just retire when F38 retires, if people want to maintain them we should encourage them to either work with fedora officially or have them do it at the compose step in timothee's gitlab namespace. What do we think about that?

bsherman:
> IMO let's name them what upstream Fedora names them, and then for things that timothee adds to the repo we follow we should just mirror as a community service.

This is what I expected, and I agree.

The only nuance here is really the naming. I'm not 100% sure what Fedora will name their images. Right now their "official" name for the silverblue image is `fedora-silverblue` and we're going to keep `silverblue` so... already we've departed a bit.

So what I'm really asking is are we going to name ours `sway-atomic-main` and `budgie-atomic-main` or just `sway-main` and `budgie-main`.

And... similarly, if we are going to do that for those, are we going to stick with `lazurite` and `vauxite` or are going to proactively switch to `lxqt-main` and `xfce-main`?

> Things that we don't have maintainer for like mate (and other repos in the org) we should just retire when F38 retires, if people want to maintain them we should encourage them to either work with fedora officially or have them do it at the compose step in timothee's gitlab namespace. What do we think about that?

I agree with this point. We'll be retiring some things in our org which don't really get maintained.

castrojo:
I'm seeing timothee at kubecon I'll ask about the image names and get a list

dylanmtaylor:
I think this looks good to go in to get us some initial  F40 images

noelmiller:
Stick with the names being used by upstream in my opinion. If they change, we can do a rename at that time. @castrojo if you can get something confirmed by Timothee, we would appreciate it!

bsherman:
> Stick with the names being used by upstream in my opinion. If they change, we can do a rename at that time.

I agree, this is certainly the simplest approach and doesn't require guesswork.

Even with upstream branding name changes, so far the image names haven't actually changed, so maybe they won't.

It's probably worth having this now on current names, and renaming images if needed later.

bsherman:
@travier  -  I was told you can provide information on the upcoming naming of Atomic Desktop images?

dylanmtaylor:
> @travier - I was told you can provide information on the upcoming naming of Atomic Desktop images?

Why don't we merge in with the current names so it is consistent and then also double-tag the layers with the new names if the image names change?

bsherman:
Ok, l'm willing convert to this to an open PR now, with the caveat that we VERY LIKELY will be renaming Fedora 40 images as we see changes from upstream in the next few weeks.

@ublue-os/approver

dylanmtaylor:
> Ok, l'm willing convert to this to an open PR now, with the caveat that we VERY LIKELY will be renaming Fedora 40 images as we see changes from upstream in the next few weeks.
> 
> @ublue-os/approver 

Would it not make sense to have the same release under two different names, the old ones and the new ones? That way it would be the least confusing to users.

castrojo:
How would two sets of names be less confusing? This feels like one of those cases where we're stuck with fedora's chosen name schemes?

dylanmtaylor:
> How would two sets of names be less confusing? This feels like one of those cases where we're stuck with fedora's chosen name schemes?

People already using the current name would see a new version with valid updates on it and not have to rebase.

castrojo:
Yeah not sure what we can do about that, waiting for the fedora names/images was one of the conditions of us leaving beta, and we're not going to bring the old tech debt with us for release.

This is one of the reasons why we don't recommend direct usage of these images. Downstreams who aren't paying attention will notice when their builds stop updating and update, and we special cased the one offs last cycle.

And the people who rebased know how to do it already so it's probably not too much of an ask for them to do it again considering that's really the only way to end up on these images anyway. Though we'll make sure we document it in the release notes.

dylanmtaylor:
> Yeah not sure what we can do about that, waiting for the fedora names/images was one of the conditions of us leaving beta, and we're not going to bring the old tech debt with us for release.
> 
> This is one of the reasons why we don't recommend direct usage of these images. Downstreams who aren't paying attention will notice when their builds stop updating and update, and we special cased the one offs last cycle.
> 
> And the people who rebased know how to do it already so it's probably not too much of an ask for them to do it again considering that's really the only way to end up on these images anyway. Though we'll make sure we document it in the release notes.

I'm just thinking having two image names w/ the same layers wouldn't be too hard. It's the equivalent of having :40 and :latest IMO.

p5:
But where would we draw the line?  Do we removed them in 41? 42? 
I'd say sticking with upstream's choices and timelines is the best option for us IMO.

dylanmtaylor:
> But where would we draw the line? Do we remove them in 41? 42? I'd say sticking with upstream's choices and timelines is the best option for us IMO.

Do we have to ever remove them? Adding an alias is basically free, just do a copy/push to both names in GHCR after the build.

p5 (approved):
Let's just merge this to get Kinoite and Bluefin unblocked, and decide what to name the images when Fedora changes the names of the images they publish.

I think we're pretty sure we WILL be using the new names, whatever they may be, so the question that remain (and can be answered later) is:
1. Do we maintain the old image names (e.g. onyx)?

dylanmtaylor:
Odd that 40 is failing now. It worked on commit https://github.com/ublue-os/main/pull/521/commits/77631cacd4b2548439be22d80cc3fb6ffc144400

castrojo (approved):
The images affected each have a fraction of the dl's of the other images, and onyx is dead last in all DLs, so chalking those up to CI noise and churn. Not blocking the release for that, let's goooooooooooooo.

travier:
So, the idea for F40 was to avoid doing painful renames and only change visible names, so we'll likely keep the Sericea and Onyx names for F40. For future releases or new artifacts (the container images built from fedora infra for example), we can make the change so we are probably going to move to `sway-atomic` and `budgie-atomic`.

I'll ping you when we make the change or decide on the new names.

bsherman:
> so we'll likely keep the Sericea and Onyx names for F40. For future releases or new artifacts (the container images built from Fedora infra for example), we can make the change so we are probably going to move to `sway-atomic` and `budgie-atomic`.

This seems reasonable to me. 

> I'll ping you when we make the change or decide on the new names.

Thank you!

PR #524, opened 2024-03-23
feat(ci): Initial support for aarch64

EyeCantCU:
### Guidelines
In this project we follow [Semantic PRs][1] and [Conventional Commits][2] to describe features and fixes we made. It would be nice if you did too. Not sure what this all means? Here are some [examples][3]. Did you already use this? Awesome, thanks again!

[1]: https://github.com/Ezard/semantic-prs
[2]: https://www.conventionalcommits.org/en/v1.0.0/#summary
[3]: https://www.conventionalcommits.org/en/v1.0.0/#examples


.github/workflows/build.yml
```diff
@@ -193,16 +217,17 @@ jobs:
           command: |
             # pull the base image used for FROM in containerfile so
             # we can retry on that unfortunately common failure case
-            podman pull quay.io/${{ env.SOURCE_ORG }}/${{ env.SOURCE_IMAGE }}:${{ matrix.major_version }}
+            docker pull --platform linux/${{ matrix.build_arch }} quay.io/${{ env.SOURCE_ORG }}/${{ env.SOURCE_IMAGE }}:${{ matrix.major_version }}
```
dylanmtaylor, line 220: Why are we switching from podman -> docker?

EyeCantCU: We aren't. Just testing because behavior differs here. This PR is a draft

FlyinPancake:
Went through some logs, and I found that the `aarch64 ` images are tagged as `arm64` on `quay.io`

castrojo:
This is way out of date, we're solving this another way, thanks anyway!

PR #525, opened 2024-03-24
chore(ci): split build workflows by fedora version

bsherman:
This will allow us to see if a specific version of Fedora is building or not, plus allow rebuilding only a specific version.

Extra bonus: this will let us adjust merge checks to require success of a subset of versions, eg
"Check all 38 builds successful" and "Check all 39 builds successful" but not 40.

This is nice for pre-release builds like what we have now.


bsherman:
This won't actually auto-merge, since I've changed the builds to run version specific "Check all NN builds successful" jobs instead of the single "Check all builds successful".

Once we get approvals, i'm happy to merge, or someone else can.

dylanmtaylor:
Can this be used to remove a lot of the version-conditional logic in the reusable part by moving it into the version-specific files?

EyeCantCU:
> Can this be used to remove a lot of the version-conditional logic in the reusable part by moving it into the version-specific files?

That's a great idea, but unfortunately, with how inheritance works in this instance, that's easier said than done and could lead to a bunch of redundant code

bsherman:
> Can this be used to remove a lot of the version-conditional logic in the reusable part by moving it into the version-specific files?

As @EyeCantCU, I don't think it would help much in this case, and would probably cause other code duplication.

One thing I did contemplate moving, the conditionals for latest, stable and gts tagging... However, i chose to move them into the tag generation step (the only place they are used) and keep it there instead of passing in as inputs, because this keeps all the logic for it together, making it easier to read the whole context in one place.

EyeCantCU (approved):

.github/workflows/build-39.yml
```diff
@@ -0,0 +1,15 @@
+name: ublue main 39
+on:
+  pull_request:
+  merge_group:
+  schedule:
+    - cron: '0 5 * * *'  # 3pm UTC everyday (timed against official fedora container pushes)
```
noelmiller, line 6: Should this be set to 15?

.github/workflows/build-40.yml
```diff
@@ -0,0 +1,15 @@
+name: ublue main 40
+on:
+  pull_request:
+  merge_group:
+  schedule:
+    - cron: '0 10 * * *'  # 3pm UTC everyday (timed against official fedora container pushes)
```
noelmiller, line 6: This is set to 10?

bsherman: That's a fever error. I'll correct.

.github/workflows/reusable-build.yml
```diff
@@ -107,7 +96,21 @@ jobs:
...
           # Generate a timestamp for creating an image version history
           TIMESTAMP="$(date +%Y%m%d)"
-          VARIANT="${{ matrix.major_version }}"
+          VARIANT="${{ matrix.fedora_version }}"
+
+          if [[ "${{ matrix.fedora_version }}" -ge "38" ]]; then
+            IS_LATEST_VERSION=false
+            IS_STABLE_VERSION=true
+            IS_GTS_VERSION=true
+          elif [[ "${{ matrix.fedora_version }}" -ge "38" ]]; then
```
noelmiller, line 105: Should this be 38 again?

noelmiller (approved):

Issue #369, opened 2023-09-25
Proposal: reduce the scope of ublue-os/main images


bsherman:
## Summary
Revise scope of [Universal Blue Main Images](https://universal-blue.org/scope/#main-images) such that kmods (and associated kernel packages) are no longer included.

## Description
The purpose of this change is to streamline build processes(current and future), maximize use of `*-main` images as a foundation for downstream efforts, and enable maximal flexibility for users running these images.

The primary change removes the following packages from `*-main` images:
  - kernel-devel-matched
  - kmods
      - xpadneo
      - xpad-noone
      - xone
      - openrazer
      - v4l2loopback
      - wl

## Scope of Changes
- Main (repo)
    - Remove above packages from `*-main` builds
    - Stop building `*-nokmods` as those images will be identical to `*-main` after this change
- ASUS/Surface
    - Build directly from `*-main`
- Website
    - Clarify that `*-main` images should not include kmods, kernel-devel, kernel-tools packages
    - Emphasize that `*-main` may be used directly but is primarily a foundational part of the Universal Blue toolkit.
    - Better document use of `akmods`.
    - Announcement of change on Blog/Github/Discord

### Not in scope
- Removal of `*-nvidia` images: they MUST have the nvidia kmod, so that will stay (still based on `*-main` so will lose the other kmods)

## Benefits to Universal Blue
- **Reduce complexity of builds and number of images maintained**
- **Provide a single foundational image for all downstream Universal Blue images/users**
- Prevent artificial restrictions on downstream usage *(kernel specific packages installed in `*-main` cannot be removed downstream or on a running system which prevents use as a base for images like asus/surface, any changes to kernel, as well as layering of related packages on running systems)*

## Feedback
A conversation in [discord](https://discord.com/channels/1072614816579063828/1072617059265032342/1155886903304925185) was had about narrowing the scope of our `*-main` images such that they no longer include kmods. 

There were questions about impact to users of existing `*-main` images, the primary concern being removal of `xpadneo` and `v4l2loopback` as they are the two kmods which have existed for a longer time. 

It was suggested we suggest users use `*-nokmods` images if they have issues with kmods in `*-main`, but that does not address the desire to simplify build processes within the Universal Blue organization.

### Ways to address 
We can suggest use of bazzite, bazzite-gnome or bluefin for users not wishing to build a custom image.

In the future, to partially offset the impact of these changes, `just` scripts may be implemented that simplify the installation of the kmods Universal Blue provides.

## Timeline

[...]

Issue #91, opened 2023-03-30
Request apr and apr-util be included for Davinci Resolve


Zindswini:
### Describe the package

[Davinci Resolve](https://www.blackmagicdesign.com/products/davinciresolve), a popular video editing/color suite requires two layered packages, `apr` and `apr-util`, but otherwise works out of the box on Silverblue using the official installer. It conveniently installs to /opt, which is tracked by ostree. (Lucky!)

These are small packages, and I believe that including them would streamline the linux experience for someone trying to follow the steps on the Resolve website.

If these are out of scope, I would instead like to add an entry to the website.

```
Dependencies resolved.
================================================================================================================================================
 Package                                 Architecture                  Version                              Repository                     Size
================================================================================================================================================
Installing:
 apr                                     x86_64                        1.7.0-16.fc36                        fedora                        124 k
 apr-util                                x86_64                        1.6.1-20.fc36                        fedora                         95 k
Installing weak dependencies:
 apr-util-bdb                            x86_64                        1.6.1-20.fc36                        fedora                         12 k
 apr-util-openssl                        x86_64                        1.6.1-20.fc36                        fedora                         15 k

Transaction Summary
================================================================================================================================================
Install  4 Packages

Total download size: 246 k
Installed size: 541 k
```

Thanks!

### Image

All Images

castrojo:
Thanks for investigating this, is there anything special we should add on the website? Is the installer an RPM or a script? It might be worth a small set of instructions on the website if it's nonobvious.

Zindswini:
The download page gives a `.run` executable which when ran opens an installer that eventually extracts the files into `/opt/resolve` after accepting terms. The installation of the main app worked fine, nothing in the installer seemed wrong besides a small theming error. 

I did encounter a bit of an issue, the `.desktop` files weren't put in place. There are placeholder files in `/opt/resolve/share/` which have "RESOLVE_INSTALL_LOCATION" instead of "/opt/resolve". I had to manually replace those and then copy them into `~/.local/share/applications`. This step could probably be simplified to a one-liner and put on the website.

Blackmagic also makes some hardware which the logs mention have udev rules but I don't really know how udev rules work and I don't have any of the devices to test with. FWIW the logs didn't mention that they failed to install. I also can't test the AMD side of this so if someone wants to give that a shot that might be helpful.

Overall though once I figured out how to fix the desktop file everything ran perfectly including (nvidia) hardware encoding. I was very impressed, everything "just works" for my specific configuration.

[...]

PR #544, opened 2024-04-06
feat: reduce package scope and remove redundancies

qoijjj:
To be lazy consensus voted

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4,20 +4,16 @@
             "all": [
                 "alsa-firmware",
                 "android-udev-rules",
-                "apr",
-                "apr-util",
                 "distrobox",
                 "ffmpeg",
                 "ffmpeg-libs",
                 "ffmpegthumbnailer",
-                "fzf",
                 "google-noto-sans-balinese-fonts",
                 "google-noto-sans-cjk-fonts",
                 "google-noto-sans-javanese-fonts",
                 "google-noto-sans-sundanese-fonts",
                 "grub2-tools-extra",
                 "heif-pixbuf-loader",
-                "htop",
                 "intel-media-driver",
                 "just",
                 "kernel-tools",
@@ -41,17 +37,13 @@
                 "smartmontools",
                 "solaar-udev",
                 "symlinks",
-                "tcpdump",
-                "tmux",
                 "traceroute",
                 "vim",
                 "wireguard-tools",
                 "zstd"
             ],
             "silverblue": [
-                "adw-gtk3-theme",
                 "gnome-epub-thumbnailer",
-                "gnome-tweaks",
                 "gvfs-nfs"
             ],
             "kinoite": [
@@ -194,7 +186,6 @@
                 "xrdb"
             ],
             "sericea": [
-                "clipman",
                 "gvfs-mtp",
                 "thunar-volman",
                 "tumbler"
--- a/packages.json
+++ b/packages.json
@@ -8,6 +8,7 @@
                 "ffmpeg",
                 "ffmpeg-libs",
                 "ffmpegthumbnailer",
+                "fzf",
                 "google-noto-sans-balinese-fonts",
                 "google-noto-sans-cjk-fonts",
                 "google-noto-sans-javanese-fonts",
--- a/packages.json
+++ b/packages.json
@@ -37,13 +37,16 @@
                 "smartmontools",
                 "solaar-udev",
                 "symlinks",
+                "tcpdump",
                 "traceroute",
                 "vim",
                 "wireguard-tools",
                 "zstd"
             ],
             "silverblue": [
+                "adw-gtk3-theme",
                 "gnome-epub-thumbnailer",
+                "gnome-tweaks",
                 "gvfs-nfs"
             ],
             "kinoite": [


packages.json
```diff
@@ -4,20 +4,16 @@
             "all": [
                 "alsa-firmware",
                 "android-udev-rules",
-                "apr",
-                "apr-util",
                 "distrobox",
                 "ffmpeg",
                 "ffmpeg-libs",
                 "ffmpegthumbnailer",
-                "fzf",
```
castrojo, line 13: We need this one for the just selectors in the terminal.

castrojo (changes requested):

fiftydinar (changes requested):
I suggested to remove `nvtop` too, since Mission Center exists, which uses this as a backend for GPU information.

packages.json
fiftydinar: I believe that generic bash dialog is used as a fallback if `fzf` & `gum` are not installed.
There is maybe some degradation in looks when using bash selector instead of `fzf` one, but it's still a viable solution I think.

So I also think that this one can be removed.

For final decision, I would consult @HikariKnight

bsherman: I'm personally indifferent on this package, but it was added over a year ago ( #84 ) to assist with `ujust` recipes. I am inclined to keep it without a strong reason to remove.

packages.json
```diff
@@ -4,20 +4,16 @@
...
                 "google-noto-sans-balinese-fonts",
                 "google-noto-sans-cjk-fonts",
                 "google-noto-sans-javanese-fonts",
                 "google-noto-sans-sundanese-fonts",
                 "grub2-tools-extra",
                 "heif-pixbuf-loader",
-                "htop",
```
bsherman, line 20: This was added over a year ago ( #82 ) as an opinionated, lightweight helpful utility. The only concern I've ever seen about this is that it comes with a desktop file which shows up in DE app lists and some people seem to find that offensive.

I'm not inclined to remove it.

fiftydinar: > The only concern I've ever seen about this is that it comes with a desktop file which shows up in DE app lists and some people seem to find that offensive.

This one can be easy solved with:
echo 'Hidden=true' >> /usr/share/applications/htop.desktop

packages.json
```diff
@@ -42,7 +38,6 @@
                 "solaar-udev",
                 "symlinks",
                 "tcpdump",
-                "tmux",
```
bsherman, line 45: This was added nearly a year ago ( #190 ) due to multiple contributors/users who believed it to be a useful, lightweight tool which is valuable on the host main images and who realized it was being added on several downstream images.

I'm not inclined to remove it.

packages.json
bsherman: Yep, I am fine with hiding the desktop icon if the group wishes to do that.

fiftydinar: > Yep, I am fine with hiding the desktop icon if the group wishes to do that.

I would also do that for `nvtop` too, if majority agrees.

packages.json
```diff
@@ -194,7 +189,6 @@
                 "xrdb"
             ],
             "sericea": [
-                "clipman",
```
bsherman, line 197: This was added in #273 and it does fill a gap in the upstream Fedora Atomic Sway image. Unless this is provided upstream, I am not inclined to remove this.

packages.json
castrojo: I'm -1 for mangling .desktop files in main - I agree it's annoying, which is why we do it downstream. 😄

bsherman: Also agreed, though, I'd suggest that be a distinct PR.

packages.json
qoijjj: @bsherman upstream already provides wl-clipboard

packages.json
```diff
@@ -4,20 +4,16 @@
             "all": [
                 "alsa-firmware",
                 "android-udev-rules",
-                "apr",
-                "apr-util",
```
bsherman, line 8: APR was added specifically to support installations of Davinci Resolve ( #91 ) and was a nice example of a user providing a clean request for package support or alternative documentations. 

There's no harm to keeping this, but removing it is reversing a past decision to save... 246k on a download? 

I'm not sure I would support adding this today, but I'm not really in favor of removing support without a strong reason.

packages.json
qoijjj: @bsherman ultimately this PR is about giving `main` a well-defined scope, which it currently lacks. Without which, it is not clear at all what belongs here, upstream, or downstream.

If these packages are universally useful, they should be added in fedora comps.

packages.json
qoijjj: > save... 246k on a download?

Image size is not the purpose of this PR :slightly_smiling_face:

packages.json
noelmiller: Second this. Similar to vim this is a popular tool for folks and adds almost no size to the image to include it.

bsherman:
I probably could have just given a single comment, but I was initially trying to take each package removal on its own and give feedback.

Over time as contributors we've tightened the scope of main, as a result we became much more strict about what packages were ADDED to the main images. We also did remove some packages which were previously added, but usually removals of previously added packages were reserved for packages which caused some issue downstream. 

Example, we removed kmods from the main images (see #369 ) because they prevented users and dowstream images from using alternative kernels. That was a a conflict which arguably reduced the main images' ability to support hardware (an in scope goal), but it was important for the main images to be useful as a common base.

So, despite tightened scope, which, if applied today may prevent adding some of packages which are currently included, if those pacakges don't cause downstream problems, our position up to now has been to be cautious about removals.

packages.json
bsherman: @qoijjj I just addressed this in top-level comment.

fiftydinar:
I think it's good to sort-out what should be removed indefinitely & what should be proposed to integrate into Fedora directly, so we can remove that package later, without making any regressions to the user.

qoijjj:
> So, despite tightened scope, which, if applied today may prevent adding some of packages which are currently included, if those pacakges don't cause downstream problems, our position up to now has been to be cautious about removals.

@bsherman I understand that, but it's confusing to contributors and users alike when there are say two packages, one included in ublue main and one excluded from it, yet both are out of scope on paper. It means that the true scope is functionally an everchanging and subjective matter, as opposed to an open and objective matter. While having a subjective and changing scope is one way to do things, it is fairly limiting.

Ideally this repo would have a clear line in the sand of what belongs in it, vs what should go upstream and what should go downstream. This line should be documented, enforced, and made clear to users and contributors, so that no one needs to ask whether something is in scope or not. It should be clear from the documentation and the existing changeset. As of today, this is not possible.

xynydev:
I think it would be great if `package.json` allowed comments in the case that potentially out-of-scope packages are kept due to continuity. It could explain the reasoning behind keeping those packages and rejecting new ones. Sadly I don't think there's a comment-compatible parser command line config available by default on Fedora.

But anyway. The most important discussion here boils down to QoL package scope. Do we...?
- only include codecs and hardware enablement
   - so no packages that Fedora could ship
   - and what does this mean wrt `ujust`
- include codecs, hardware enablement, and packages / tweaks that...
   - need to be installed on the image or layered
   - are generally useful, expected, or required for a functional configuration
   - potentially stuff like `adw-gtk3`
- include codecs, hardware enablement, and packages /tweaks that...
   - users might reasonably expect to be there
   - don't take too much disk space
   - fix problems without creating them
   - potentially stuff like `vim` and `htop`
- include codecs, hardware enablement, and packages / tweaks that...
   - improve user experience in specific use cases
   - might be opionionated by nature
   - are things we find useful and suggest for users
   - are maybe things that Fedora should be doing upstream, but isn't
   - potentially stuff like `distrobox` and `apr`

Of course, this is an imperfect spectrum, and a lot of these things can overlap. But, as our goal is to refine the scope, I want to make sure we that think of the general scope 'rules' and their implications first, before getting too deep into how changes to the actual package set should be rolled out.

qoijjj:
@xynydev 

I agree with you. The purpose of this PR isn't to just tear out right away everything that isn't in scope. This was supposed to be the first of several, including only the least controversial removals

I agree that adding comments to the out of scope packages or something to that effect could help identify what needs to be removed long-term to bring this repo more in line with the "line in the sand" scope.

> and what does this mean wrt ujust

ujust along with ujust/config provides hardware enablement functionality like for optimus laptops. It should fall under hardware enablement for that reason

> are generally useful, expected, or required for a functional configuration

This should be upstream

> users might reasonably expect to be there

upstream

> don't take too much disk space

way too broad, this would mean installing every package under 20MB :smile: 

> fix problems without creating them

Since packages added upstream cannot be removed downstream at this time, every out of scope package that's added adds a small downstream user inconvenience

> potentially stuff like vim and htop

They should be added upstream if they have universal utility

> might be opionionated by nature
> are things we find useful and suggest for users
> are maybe things that Fedora should be doing upstream, but isn't

these would belong downstream

castrojo:
> It could explain the reasoning behind keeping those packages and rejecting new ones.

Most of the packages have associated issues that have been filed in the repo with a justification, etc.

fiftydinar:
@qoijjj What if upstream doesn't accept the package that belongs to them in your view? Should main image keep that package, or not ship it at all?

qoijjj:
@fiftydinar If a package has universal utility and can be shipped by Fedora (i.e. no non-free stuff) then I would think we keep it here until upstream makes a decision and leave it in place with a link to the upstream issue if they say "no".

But each out of scope package should have a discussion with upstream linked to it so we can keep track of why/why not upstream is/isn't pulling in packages that are in scope upstream.

qoijjj:
also @xynydev one other note is that json sadly doesn't support comments :smiling_face_with_tear:

xynydev:
> ujust along with ujust/config provides hardware enablement functionality like for optimus laptops. It should fall under hardware enablement for that reason

I counted 27 non-duplicate upstream `ujust` recipes on my system. ~20 of them I would not count as hardware enablement, ex.:
- `bios                           # Boot into this device's BIOS/UEFI screen`
- `distrobox-assemble CONTAINER="prompt" ACTION="create" FILE="/etc/distrobox/distrobox.ini" # Create distroboxes from a defined manifest`
- `distrobox-new IMAGE="prompt" NAME="prompt" HOMEDIR="" # Create a new custom distrobox`
- `install-brew                   # Install Homebrew | https://brew.sh`
- `install-obs-studio-portable    # Install obs-studio-portable from wimpysworld, which bundles an extensive collection of 3rd party plugins`
- `logs-last-boot                 # Show all messages from last boot` 

Obviously, many of these are useful, but I don't know how they fit into this scope:

> Focus on things that Fedora CANNOT ship:
> Hardware enablement.
> Non-free codecs aka. “a bunch of that RPMFusion stuff”.
> Built in Nvidia drivers with multiple version support.
> 
> We should be brutal about making things out of scope.

fiftydinar:
@xynydev
I think that ujusts is software enablement. It allows you to do useful things, without having to search what each command does. It's like simplified Yast, but for Fedora & in TUI.

It's most useful for easy troubleshooting imo.

I think that ujust is different in this regard, compared to classic bling packages, which affect downstream.
Ujust modifications by Universal Blue can be easily removed if desired, it doesn't harm downstream as out-of-scope packages do.

But it is indeed useful to know what should go to ujust & what shouldn't.
So far, all scripts are relatively good supported & I find them great.

qoijjj:
closing out old PRs

PR #558, opened 2024-04-15
feat: Add clevis-dracut to enable tang decryption

m2Giles:
Clevis is another scheme for supporting LUKS decryption via tpm2 and/or a network based decryption scheme. Clevis is already included in our images; however, clevis-dracut is not. We've recently enabled initramfs generation in our downstream builds to support using tpm2, fido2, and pkcs11. Adding clevis-dracut will enable tang as well.

Additionally, clevis-dracut is already included in coreOS.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -6,6 +6,7 @@
                 "android-udev-rules",
                 "apr",
                 "apr-util",
+                "clevis-dracut",
                 "distrobox",
                 "ffmpeg",
                 "ffmpeg-libs",


noelmiller:
I would like to get others opinions, but I think this falls in the scope of main. Improves base functionality for folks and specifically is useful in a home lab setting. 
@ublue-os/approver

bsherman:
I will make my obligatory cautionary statement.

Anything added is hard to remove later, so let's be very confident before adding things to `main`, even things that seem like obvious wins.

With that said, I'm not opposed, but I do think we should be responsible and answer a few questions before adding anything to `main`.

Are there ANY potential downsides? 

Does `clevis-dracut` pull in any other dependencies?  (given clevis and dracut already are in our images, I'd be surprised, but I still want to know)

Under what category does this fit [per our scope docs](https://universal-blue.discourse.group/docs?topic=51#main-images-1)?

m2Giles:
This was considered upstream for silverblue.

https://github.com/fedora-silverblue/issue-tracker/issues/409

However, it seems specifically aimed at the tpm2 use-cas

bsherman:
> This was considered upstream for silverblue.

We can definitely do things upstream does not. That's why we exist. So no need to justify that.
 
> I will investigate what this could pull in as well.

Thank you!  I really think we just need to answer some standard questions (at least the 3 I asked) before merging any new package adds. That way we'll at least have documented what expected changes are which we all agree to.

m2Giles:
More Thoughts:
Beyond LUKS decryption, this can be used to decrypt arbitrary secrets bound by either tpm2, tang, or combination. This package enables us to do that inside of the initramfs. I've personally used this for for storing decryption secrets for non LUKS encrypted devices. Systemd has mechanisms for storing secrets in /var but this has flexibility.

Downsides:

1. This package is extremely unlikely to be used. LUKS is already underutilized. This is adding an additional decryption method. The user who already knows about using Tang, says it meets their requirements, would likely have little issue layering this package.

2. This package requires additional configuration to work. Downstream we generate initramfs inside the build and can add the dracut module, but we do not currently do this in main. Though, adding this as a build step is not out of the question. Even with the initramfs generated, either kargs or additional configs inside the initramfs are necessary for configuring networking in early boot.

How Does This Fit:
Quality of Life. Hardware Enablement.

This package would need to be layered in order to utilize. It enables systems without TPMs to do automatic decryption with communication to a tang server. For systems with tpm2, it can still setup automatic decryption without needing secureboot configured (steamdeck has no keys, users that don't want to use secureboot). Clevis and Tang is a supported (and often recommended) solution for setting up systems to decryption in automatic deployments since it does not require a valid secureboot state. CoreOS has this as supported method as part of ignition.

Unlike a tpm decryption scheme, this form of automatic decryption is usually tied to an internal network. Meaning that automatic decryption only occurs in select environments.

Additionally, clevis supports SSS as a pin which can combine multiple other secrets meaning that you can combine having both tpm2 and Tang decryption before falling back to password prompt or whatever. There is also a third party yubikey support.

Packages this Pulls in:
TODO

m2Giles:
On base - image with no DE:
```
Installing 10 packages:
  clevis-20-1.fc39.x86_64 (updates)
  clevis-dracut-20-1.fc39.x86_64 (updates)
  clevis-luks-20-1.fc39.x86_64 (updates)
  clevis-pin-tpm2-0.5.3-2.fc39.x86_64 (updates)
  clevis-systemd-20-1.fc39.x86_64 (updates)
  dracut-network-059-16.fc39.x86_64 (updates)
  jose-11-8.fc39.x86_64 (fedora)
  libjose-11-8.fc39.x86_64 (fedora)
  libluksmeta-9-16.fc39.x86_64 (fedora)
  luksmeta-9-16.fc39.x86_64 (fedora)
```

The same packages are installed on silverblue-main indicating that something else pulls in clevis in our downstream builds.

bsherman:
With the recent addition of https://github.com/ublue-os/main/issues/593 I think we can close this PR since all ublue-os images have the appropriate dracut configs to meet the main goal of this PR. Plus, all hwe, bazzite, and bluefin images are building the initramfs which enables the LUKS unlock options.

Issue #527, opened 2024-03-24
F38 EOL: remove F38 builds and version specific workarounds


bsherman:
Mostly this should be removing "version 38" or the `build-38.yml` workflows from appropriate locations.

But also there's a bunch of custom version 38 stuff for kmods in main, etc which needs to be removed.

PR #562, opened 2024-04-22
chore: remove Fedora 38 & kmods images

p5:
Remove kmods builds
Remove Fedora 38

### Do not merge until a few weeks after Fedora 40 dropped

Relates: #527


.github/workflows/reusable-build.yml
```diff
@@ -96,15 +65,15 @@ jobs:
...
             IS_LATEST_VERSION=false
             IS_STABLE_VERSION=true
             IS_GTS_VERSION=true
-          elif [[ "${{ matrix.fedora_version }}" -eq "39" ]]; then
+          elif [[ "${{ matrix.fedora_version }}" -eq "40" ]]; then
             IS_LATEST_VERSION=true
             IS_STABLE_VERSION=true
             IS_GTS_VERSION=false
-          elif [[ "${{ matrix.fedora_version }}" -eq "40" ]]; then
+          elif [[ "${{ matrix.fedora_version }}" -eq "41" ]]; then
```
p5, line 76: I know we don't really need this, but no point ripping this out when it'll be added in 4 months again.

storopoli:
Is this for the Fedora 40 release?
Then `latest` will de F40 and `gts` F39?

bsherman:
@p5 I don't think we should tie the release/promotion of F40 to stable/latest to removing F38 builds. Shouldn't removing F38 be a distinct act?

As much as I want to stop maintaining 38, I really am hesitant to drop those builds this week... feels like we should get a bit closer to upstream's EOL first.

bsherman:
>  `latest` will be F40 and `gts` F39?

yes

p5:
> @p5 I don't think we should tie the release/promotion of F40 to stable/latest to removing F38 builds. Shouldn't removing F38 be a distinct act?
> 
> As much as I want to stop maintaining 38, I really am hesitant to drop those builds this week... feels like we should get a bit closer to upstream's EOL first.

Understood.  So including only the Containerfile ARG defaults and 
```
          if [[ "${{ matrix.fedora_version }}" -eq "39" ]]; then
            IS_LATEST_VERSION=false
            IS_STABLE_VERSION=true
            IS_GTS_VERSION=true
          elif [[ "${{ matrix.fedora_version }}" -eq "40" ]]; then
            IS_LATEST_VERSION=true
            IS_STABLE_VERSION=true
            IS_GTS_VERSION=false
          elif [[ "${{ matrix.fedora_version }}" -eq "41" ]]; then
```
is enough for tomorrow?

p5:
Have repurposed this PR to only deprecate Fedora 38 and kmod builds.  Will bump Fedora 39 and 40 in [this PR](https://github.com/ublue-os/main/pull/563)

bsherman:
@p5 I would suggest that "deprecate" means "not recommended for use" but this PR will remove 38 so it cannot be used.

dylanmtaylor (changes requested):
You should delete lxqt from image_name

p5:
> You should delete lxqt from the image_name list.

Good catch
Done in [9508130](https://github.com/ublue-os/main/pull/562/commits/95081301c6f135815411a4a567849cb34e76cad6)

bsherman:
> Remove kmods builds Remove Fedora 38
> ### Do not merge until a few weeks after Fedora 40 dropped

Regarding the "when" of removing Fedora 38... we recently updated our governance document to make this more clear:

> Our goal is to build two currently supported versions of Fedora, the most recent and one previous version.

https://universal-blue.discourse.group/docs?topic=51

Given this update, we should be good to remove immediately.

bsherman (approved):

KyleGospo (approved):

PR #563, opened 2024-04-22
feat: bump Fedora 40 to latest and Fedora 39 to stable

p5:
Bump Fedora 40 to latest
Bump Fedora 39 to stable
DO NOT TOUCH BSHERMAN'S BELOVED FEDORA 38 OR KMODS

I am keeping Fedora 41 in the condition because someone will be re-adding it back in 4 months anyway.

## DO NOT MERGE UNTIL 23RD APRIL


.github/workflows/reusable-build.yml
bsherman, line 1: I didn't catch this properly in the draft.

for this PR, we should retain 38, 39, 40 but set the correct flags for each.

38: stable, not gts, not latest
39: stable, gts, not latest 
40, stable, latest, not gts

bsherman (changes requested):

.github/workflows/reusable-build.yml
bsherman: and there is no 41 yet, at least not in our builds, so no need to add it.

p5: So we had this incorrect before?  Both 38 and 39 were set to stable, so I had replicated it here.  

I chose to keep 41 because we'll be adding it back in 4 months again, but I'll remove it

bsherman: > I chose to keep 41 because we'll be adding it back in 4 months again, but I'll remove it

I don't actually mind if 41 is here early, I had missed the context from your pre-emptive comment on the other PR. :-)

p5: Sorry, so I understand - your requested change is to add back the Fedora 38 condition, with IS_STABLE_VERSION=true, IS_LATEST_VERSION=false and IS_GTS_VERSION=false?

bsherman: I'm sorry for the confusion.

38 does not cease to be stable, just because it's old.

bsherman: No worries! I didn't actually create the original logic for STABLE/LATEST, nor even GTS, but I have been part of maintaining it and obviously switched us from using matrix includes to this conditional shell snippet.

It may be that we could omit an outgoing version like 38 in the current situation, but I've always ensured the state was fully populated for versions we are building to avoid any issues with undefined vars, etc.

bsherman (approved):

castrojo (approved):

PR #571, opened 2024-05-08
fix: github release install silently fails if no urls found

bcook254:
I use this script for my own personal build and found that if the script does not find any valid rpms it will silently continue. This change creates an array and simply selects the first url (as the current script does). If there are no urls then `rpm-ostree install` should fail.


bsherman:
> I use this script for my own personal build and found that if the script does not find any valid rpms it will silently continue. This change creates an array and simply selects the first url (as the current script does). If there are no urls then `rpm-ostree install` should fail.

@bcook254 I just stumbled onto this PR as it never seemed to cross my radar before.

Appreciate the simple solution, as the author of this script, I've run into the problem myself and hadn't fixed it yet.

castrojo (approved):

bsherman (approved):
I verified this fix locally.

:+1:

PR #581, opened 2024-05-16
chore: add bootc to all images

m2Giles:

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -6,6 +6,7 @@
                 "android-udev-rules",
                 "apr",
                 "apr-util",
+		        "bootc",
                 "distrobox",
                 "ffmpeg",
                 "ffmpeg-libs",
@@ -193,9 +194,7 @@
     },
     "39": {
         "include": {
-            "all": [
-                "bootc"
-            ],
+            "all": [],
             "kinoite": [
                 "xwaylandvideobridge"
             ]
--- a/post-install.sh
+++ b/post-install.sh
@@ -2,8 +2,12 @@
 
 set -ouex pipefail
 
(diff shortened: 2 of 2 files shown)


m2Giles:
While bootc will be the future, I'm wondering if we should hold off on having these on all images while the supporting infrastructure is still being put in place.

The command to make the iso work on F40 causes it to break on F39. This can be handled via an if statement during build. But maybe a point of concern for downstream images where if it isn't rerun the iso will break.

packages.json
```diff
@@ -6,6 +6,7 @@
                 "android-udev-rules",
                 "apr",
                 "apr-util",
+		        "bootc",
```
bsherman, line 9: My nitpick, lets make the spacing on this line consistent with others. :-)

bsherman (changes requested):

bsherman:
> While bootc will be the future, I'm wondering if we should hold off on having these on all images while the supporting infrastructure is still being put in place.
> 
> The command to make the iso work on F40 causes it to break on F39. This can be handled via an if statement during build. But maybe a point of concern for downstream images where if it isn't rerun the iso will break.

I support adding `bootc` ... and I think we had this in the past, no?

Users of bootc would be experimental/early-adopters who are fine with it's explicit lack of support for local layering.
Thus, all our docs which talk about command-line stuff should continue to reference `rpm-ostree` operations. 


I don't know how ISO building has anything to do with having the binary on an image to support the bootc effort.

fiftydinar:
`bootc` is not in F41, only it's dependency `bootupd` is in it.
Just noting if this PR is still followed or considered.

PR #591, opened 2024-06-26
feat: include overpass-fonts, and fix whitespace

qoijjj:
This is a 1MB package that provides really nice fonts: https://fonts.google.com/specimen/Overpass


castrojo:
This is out of scope for main images.

PR #592, opened 2024-06-28
fix: disable gdm's wayland checks

m2Giles:
For some nvidia users, gdm's udev rules prevent the wayland session from starting.

This provides an override to the file by symlinking the /etc location to /dev/null.

Note, @p5 has experienced oddities when removing the file.


castrojo (approved):

fiftydinar:
If only Nvidia users have issue with it, then I think that this should be done in Nvidia images only, not throughout `-main`

PR #604, opened 2024-07-11
feat: Improve Containerfile / enable workflow usage of containerized buildah (tag: v1)

m0gg:
These are the first bits of my suggestions to improve the buildprocess via Containerfile.  
Before going any further, i'd like to hear some feedback whether these kind of refactorings are welcome or not.


m2Giles:
Thank you for this. One thing is that we don't build with a new enough version of podman to take advantage of heredoc. It's extremely sad and unfortunate

m0gg:
> It's extremely sad and unfortunate.
I absolutely agree.

>  looks like most of the changes are taking advantage of dockers latest syntax and things like caching.
It does run as expected on recent podman though.

May I ask for the reason for such an "old" version of podman?

castrojo:
> May I ask for the reason for such an "old" version of podman?

The ubuntu builders are 22.04, we're waiting for 24.04 builders to go GA so we can get newer podman versions, which should be sometime soon.

m0gg:
> > May I ask for the reason for such an "old" version of podman?
> 
> The ubuntu builders are 22.04, we're waiting for 24.04 builders to go GA so we can get newer podman versions, which should be sometime soon.

*Sigh* ... 

In that case, i could continue to test this stuff with gitea and ubuntu-latest runners.

While familiarising myself with the ublue projects, I got the urge to refactor some parts of the build system/scripts.  
Are you open to review/comment changes requiring the newer builders?

castrojo:
We need newer builders for zstd:chunked compression anyway, so welcome to the queue! We did opt in for a while into 24.04 but we had some builder failures so we rolled back to 22.04, but it's worth investigating again. 

Open an issue on what you'd like to refactor and slap an enhancement label on it. Then we can have the other maintainers take a look and work towards consensus. Looking forward to seeing what you have in mind!

castrojo:
Looks like the version in 24.04 is also too old for heredoc support, we should perhaps consider running with the podman OBS repo so we're always on the newest version, I think someone's proposed that before?

m0gg:
> Looks like the version in 24.04 is also too old for heredoc support

Not really too old - more like: broken. I don't know what they did to that poor thing, but it told me it's version 1.33.7.

```
runner@6562b73dd4be:/$ buildah --version
buildah version 1.33.7 (image-spec 1.1.0-rc.5, runtime-spec 1.1.0)
runner@6562b73dd4be:/$ buildah build -f- <<EOC
> FROM alpine
> RUN <<EOF
> echo I am groot
> EOF
> EOC
STEP 1/4: FROM alpine
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/shortnames.conf)
Trying to pull docker.io/library/alpine:latest...
Getting image source signatures
Copying blob ec99f8b99825 done   |
Copying config a606584aa9 done   |
Writing manifest to image destination
STEP 2/4: RUN <<EOF
STEP 3/4: echo I am groot
ERRO[0002] +(UNHANDLED LOGLEVEL) &imagebuilder.Step{Env:[]string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}, Command:"echo", Args:[]string{""}, Flags:[]string{}, Attrs:map[string]bool(nil), Message:"ECHO ", Original:"echo I am groot"}
Error: building at STEP "ECHO ": Build error: Unknown instruction: "ECHO" &imagebuilder.Step{Env:[]string{"PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"}, Command:"echo", Args:[]string{""}, Flags:[]string{}, Attrs:map[string]bool(nil), Message:"ECHO ", Original:"echo I am groot"}
```

Support for here-documents was added in 1.33.0:

```
$ podman -r run --rm -i --privileged quay.io/buildah/stable:v1.33.2 buildah build -f- <<EOC
FROM alpine
RUN <<EOF
echo I am groot
EOF
EOC
STEP 1/2: FROM alpine
Resolved "alpine" as an alias (/etc/containers/registries.conf.d/000-shortnames.conf)
Trying to pull docker.io/library/alpine:latest...
Getting image source signatures
Copying blob sha256:ec99f8b99825a742d50fb3ce173d291378a46ab54b8ef7dd75e5654e2a296e99
Copying config sha256:a606584aa9aa875552092ec9e1d62cb98d486f51f389609914039aabd9414687
Writing manifest to image destination
STEP 2/2: RUN <<EOF
time="2024-07-12T20:22:43Z" level=warning msg="Path \"/run/secrets/etc-pki-entitlement\" from \"/etc/containers/mounts.conf\" doesn't exist, skipping"
I am groot
COMMIT
Getting image source signatures
Copying blob sha256:94e5f06ff8e3d4441dc3cd8b090ff38dc911bfa8ebdb0dc28395bc98f82f983f
Copying blob sha256:47f646e0d5136a366531b0238f6d2d6af4db40b315bcd4c96b37e14fd0c59c38
Copying config sha256:6e7b5d99878d8e949c40600eb4ca47dc37e6ff3be2900dcba73049e25425de44
Writing manifest to image destination
--> 6e7b5d99878d
6e7b5d99878d8e949c40600eb4ca47dc37e6ff3be2900dcba73049e25425de44
```

m0gg:
While it might seem a little hacky, this worked well when i tested it with gitea actions and the 22.04 builder: 5ca8240f348beb9d17b32efc6a78a464c5cfcd3a

m2Giles:
That's hacky, but we do have to workaround the limitations of the builders. Letting this start to run workflows

Containerfile
```diff
@@ -1,35 +1,38 @@
...
+ARG IMAGE_NAME=silverblue
+ARG SOURCE_IMAGE=silverblue
+ARG SOURCE_ORG=fedora-ostree-desktops
 ARG BASE_IMAGE="quay.io/${SOURCE_ORG}/${SOURCE_IMAGE}"
-ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION:-40}"
+ARG FEDORA_MAJOR_VERSION=40
+
+# workaround for selinux denying direct access to mounted buildcontext
+FROM scratch as context
+COPY / /
```
bsherman, line 9: Does this workaround address the problem with `tar` failing?

.github/workflows/reusable-build.yml
```diff
@@ -167,6 +167,39 @@ jobs:
...
+            exec podman run --rm \
+              --privileged \
+              --net=host \
+              --cgroups=disabled \
+              --runtime=crun \
+              --runtime-flag=cgroup-manager=disabled \
+              --security-opt=label=disable \
+              --security-opt=seccomp=unconfined \
+              --device=/dev/fuse:rw \
+              -v /workspace:/workspace \
```
m2Giles, line 195: I'm guessing this is a difference between the gitea local runner and githubs runner. I'm assuming this is supposed to be the current directory.

Containerfile
m0gg: No, do you have a reference or example for that issue?  
Initially i tried bind mounting the buildcontext directly (by omitting `from=`), but access from within the container was denied on selinux enabled podman machine executors.

.github/workflows/reusable-build.yml
m0gg: I guess you're right. It looks like the runners home directory is the github-equivalent.

m0gg:
Okay, the latest version works in github-ci

bigpod98:
not exactly all that up to speed on heredocs but what does a build gain with using heredocs

m0gg:
Now it successfully runs on GH. Maybe someone would like to trigger the workflow again?

m0gg:
> not exactly all that up to speed on heredocs but what does a build gain with using heredocs

Ultimately, it's preference. Heredocs can be written like normal scripts do: multiple lines, `set -eu`, etc., etc.  
They are a lot more maintainable than sequences of `&& \`.

But this isn't just about heredocs.

m2Giles:
I noticed that for the mount commands there is a difference between how docker and podman handle these. Docker appears to not like nested mounts. Also docker gets angry with it called context since that's a reserved word.

While we use podman for building, keeping compatibility with docker is a plus.

Overall I really like what this is doing and especially considering that we are copying in the kernel now, reducing the size of the image by removing those COPY layers is great.

m0gg:
> I noticed that for the mount commands there is a difference between how docker and podman handle these. Docker appears to not like nested mounts. Also docker gets angry with it called context since that's a reserved word.
> 
> While we use podman for building, keeping compatibility with docker is a plus.
> 
> Overall I really like what this is doing and especially considering that we are copying in the kernel now, reducing the size of the image by removing those COPY layers is great.

Interesting. A quick `just build` did not yield any complaints or warnings on my rootless docker 25.0.2 (using buildx). @m2Giles what version are you using?

m2Giles:
docker buildx is v0.16
docker is 27.0.3

Wouldn't be surprised if it's some weird ordering

m0gg:
This latest version runs without complaints on recent docker/buildx and podman. I've already put in too many commits which are may be considered unrelated or personal preference, so please comment away. I could split out some of them if desired.

m2Giles:
Mount is definitely something I really like seeing and removes a ton of dead space.

The heredoc stuff is nicer to read, but think it is style related. Overall, I do like a lot of the stuff in this PR.

Containerfile
m2Giles: This is only for SELinux handling for copying the build context into the container.

When mounting the actual build context instead doing this scratch workaround.... selinux yells at us and doesn't let us have access to the fiel.

m2Giles (approved):
I like this a lot and gives us a good blueprint for improving containerfiles in the rest of the organization.

Containerfile
bsherman: Oops, yes, @m2Giles answered for me... I missed the question. 

This part looks good to me.

Containerfile
```diff
@@ -1,39 +1,44 @@
...
-RUN mkdir -p /var/lib/alternatives && \
-    /tmp/install.sh && \
-    /tmp/post-install.sh && \
-    mv /var/lib/alternatives /staged-alternatives && \
-    rm -rf /tmp/* /var/* && \
-    ostree container commit && \
-    mkdir -p /var/lib && mv /staged-alternatives /var/lib/alternatives && \
-    mkdir -p /tmp /var/tmp && \
-    chmod -R 1777 /tmp /var/tmp
+RUN --mount=type=bind,from=ctx,source=/,target=/buildcontext \
```
bsherman, line 24: I like this bind mount and the use of the buildah wrapper to enable heredoc. This is a nice win both on eliminating wasteful COPY layers and readability.

github-release-install.sh
```diff
@@ -42,19 +42,15 @@ fi
...
-    -r \
-    --arg arch_filter "${ARCH_FILTER}" \
-    '.assets | sort_by(.created_at) | reverse | .[] | select(.name|test($arch_filter)) | select (.name|test("rpm$")) | .browser_download_url')
-for URL in ${RPM_URLS}; do
-  # WARNING: in case of multiple matches, this only installs the first matched release
-  echo "execute: rpm-ostree install \"${URL}\""
-  rpm-ostree install "${URL}"
-  break
-done
+curl --fail --retry 5 --retry-delay 5 --retry-all-errors -sL ${API} | \
```
bsherman, line 48: I don't see a compelling reason for the changes to this file.

In general, I prefer to not make style changes without a clear reason.

Please let me know if you have a strong reason for it, otherwise I request this file be reverted.

install.sh
```diff
@@ -6,37 +6,21 @@ RELEASE="$(rpm -E %fedora)"
...
-        /tmp/kernel-rpms/kernel-core-*.rpm \
-        /tmp/kernel-rpms/kernel-modules-*.rpm
-fi
+curl -fSsLo /etc/yum.repos.d/_copr_ublue-os_staging.repo https://copr.fedorainfracloud.org/coprs/ublue-os/staging/repo/fedora-"${RELEASE}"/ublue-os-staging-fedora-"${RELEASE}".repo
+curl -fSsLo /etc/yum.repos.d/_copr_kylegospo_oversteer.repo https://copr.fedorainfracloud.org/coprs/kylegospo/oversteer/repo/fedora-"${RELEASE}"/kylegospo-oversteer-fedora-"${RELEASE}".repo
+
+{
+    printf "%s\0" "${REPO_PACKAGES[@]}"
+    find "$RPMS_DIR"/{config,akmods} -type f -name "*.rpm" -print0
+} | xargs -0 rpm-ostree install
```
bsherman, line 23: It takes effort to read the new array+xargs version of this code for installing extra repos vs the previous version being a simple multi-line install command.

I'd suggest something like:
```
rpm-ostree install \
    fedora-repos-archive \
    "$RPMS_DIR"/akmods/*.rpm \
    "$RPMS_DIR"/config/*.rpm \
    "${RPMFUSION_MIRROR_RPMS}/free/fedora/rpmfusion-free-release-${RELEASE}.noarch.rpm" \
    "${RPMFUSION_MIRROR_RPMS}/nonfree/fedora/rpmfusion-nonfree-release-${RELEASE}.noarch.rpm"
```

install.sh
```diff
@@ -59,10 +43,26 @@ if [ -n "${RPMFUSION_MIRROR}" ]; then
 fi
 
 # run common packages script
-/tmp/packages.sh
+"$BUILDCONTEXT_DIR/packages.sh"
+
+# Handle Kernel Skew with override replace
```
bsherman, line 48: The `# Handle Kernel Skew with override replace` section was moved after `# run common packages script`, but we wish to know that any kernel changes cleanly install before moving on to the common packages install.

Please return to the original sequence.

bsherman (changes requested):
I'm requesting a few small changes. My reasons are either to maintain readability of the code or to avoid changes which I don't see as needed.

Happy to discuss.

Thanks!

install.sh
m0gg: Oh, i'm sorry i failed to document the reason. I moved this to "after the common packages" because i observed dracut falling back to pigz instead of the intended zstd if/when regenerating the intiramfs.

install.sh
m0gg: While i agree with the readability point, i strongly object the usage of shell glob expansion for selecting files to be passed to commands.  

It might actually be okay in this case, but also introduces non-obviuos pitfalls:
- limited argument count (most likely irrelevant)
- ignoring dot-prefixed files (most likely irrelevant)
- ignoring files in subdirectories (could lead to confusion)

In my opinion passing a unknown, variable number of (also unknown and variable) file paths as arguments should always be done in a safe way. I would not consider shell glob expansion without a filename prefix && suffix to be safe enough.

To keep readability, i'd like to propose a solution relying on two `rpm-ostree install` calls instead:

```
rpm-ostree install \
    fedora-repos-archive \
    "${RPMFUSION_MIRROR_RPMS}/free/fedora/rpmfusion-free-release-${RELEASE}.noarch.rpm" \
    "${RPMFUSION_MIRROR_RPMS}/nonfree/fedora/rpmfusion-nonfree-release-${RELEASE}.noarch.rpm"
find "$RPMS_DIR"/{config,akmods} -type f -name "*.rpm" -print0 | xargs -0 rpm-ostree install
```

github-release-install.sh
m0gg: My reason for this change was simplicity and tidiness. Eliminating the usage of a temporary file which is only cleaned up indirectly by another script seemed like an enhancement to me.

install.sh
m2Giles: The dracut run you see doesn't actually get included into the final image when used with kernel-install for some reason. Only when we manually invoke it does it seem to get included correctly.

install.sh
m2Giles: In this case, we don't actually want to install everything from akmods. We only want to install the package ublue-os-akmods-addons and exclude everythin in the kmods subdirectory. Similarly to config, the desired rpms are all in the rpm directory.

m0gg: > In this case, we don't actually want to install everything from akmods.
Yeah, i saw that. It is the `ublue-os` subdirectory which is mounted to `akmods`. Mabye that's a little confusing.

install.sh
m0gg: I removed it from this PR anyway - it doesnt belong here.

install.sh
m0gg: I updated it to install from `"$RPMS_DIR"/{config,akmods/ublue-os}` and mount `/rpms` instead of `/rpms/ublue-os`.  
This should be less confusing.  
f559aaef127c68b8bf24ab7c8ffb55c384253572

github-release-install.sh
bsherman: > Eliminating the usage of a temporary file which is only cleaned up indirectly by another script, seemed like an enhancement to me.

Good call to clean up the temp file here rather than externally. I'd like to see that added at the end of this script rather than remove the use of the temp file. I intentionally kept the temp file rather than directly process the URL to aid in troubleshooting when problems occur.

The other changes are all fine, but it seems to be subjective improvements.

m2Giles:
One thing I'm also noticing in the build logs is that dracut no longer has access to /dev/kmsg and/or /dev/logger so dracuts output are hidden.

The other change that is catching my eye is the change from a for loop to a while loop in github installer script. I know read is preferable when working with arrays, but just trying to understand why the change was made there.

The xargs change is more readable now, bracket bash is something shellcheck indicates to use at times, but I think this is more readable.

install.sh
```diff
@@ -6,36 +6,37 @@ RELEASE="$(rpm -E %fedora)"
...
-fi
+RPMFUSION_MIRROR_RPMS="${RPMFUSION_MIRROR:-https://mirrors.rpmfusion.org}"
 
-curl -Lo /tmp/rpms/rpmfusion-free-release-"${RELEASE}".noarch.rpm "${RPMFUSION_MIRROR_RPMS}"/free/fedora/rpmfusion-free-release-"${RELEASE}".noarch.rpm
-curl -Lo /tmp/rpms/rpmfusion-nonfree-release-"${RELEASE}".noarch.rpm "${RPMFUSION_MIRROR_RPMS}"/nonfree/fedora/rpmfusion-nonfree-release-"${RELEASE}".noarch.rpm
+declare -a REPO_PACKAGES=(
+    fedora-repos-archive
+    "${RPMFUSION_MIRROR_RPMS}/free/fedora/rpmfusion-free-release-${RELEASE}.noarch.rpm"
+    "${RPMFUSION_MIRROR_RPMS}/nonfree/fedora/rpmfusion-nonfree-release-${RELEASE}.noarch.rpm"
+)
```
bsherman, line 15: With the change to two `rpm-ostree install` commands a few lines below, this array variable is now unused and should be removed.

m0gg: Of course, i overlooked that, thanks.

github-release-install.sh
m0gg: Alright. This trap should do the trick.

bsherman (approved):

m2Giles (approved):

PR #610, opened 2024-07-19
Revert "feat: Improve Containerfile / enable workflow usage of containerized buildah (tag: v1)"

m2Giles:
Reverts ublue-os/main#604

Kernel Copy is not working correctly and other downstream breakages occurring.

### DIFF
--- a/.github/workflows/reusable-build.yml
+++ b/.github/workflows/reusable-build.yml
@@ -177,39 +177,6 @@ jobs:
             io.artifacthub.package.readme-url=https://raw.githubusercontent.com/${{ github.repository }}/main/README.md
             io.artifacthub.package.logo-url=https://avatars.githubusercontent.com/u/120078124?s=200&v=4
 
-      # buildah on ubuntu-22.04 builder is pretty dated and the version on
-      # ubuntu-24.04 misbehaves when using heredocs in Containerfiles
-      # (as of 2024-07-12)
-      # this wrapper script mimics toolbx/distrobox behaviour for buildah which
-      # is invoked by redhat-actions/buildah-build@v2
-      - name: Install wrapper script for containerized buildah
-        id: fix-gh-buildah
-        shell: bash
-        run: |
-          cat > /usr/local/bin/buildah <<'EOF'
-            #!/bin/sh
-            set -eux
-
-            # get local graphroot at runtime
-            GRAPH_ROOT="$(/usr/bin/buildah info | jq -r '.store.GraphRoot')"
-
-            exec podman run --rm \
-              --privileged \
-              --net=host \
-              --cgroups=disabled \
-              --runtime=crun \
-              --runtime-flag=cgroup-manager=disabled \
-              --security-opt=label=disable \
-              --security-opt=seccomp=unconfined \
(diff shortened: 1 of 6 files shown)


bsherman (approved):

bsherman:
Current failures are due to https://github.com/ublue-os/main/actions/runs/10013078525/job/27679920233#step:11:267
```
error: Could not depsolve transaction; 1 problem detected:
 Problem: package mesa-va-drivers-freeworld-24.0.5-1.fc40.x86_64 from rpmfusion-free requires mesa-filesystem(x86-64) = 24.0.5, but none of the providers can be installed
  - cannot install both mesa-filesystem-24.0.5-1.fc40.x86_64 from fedora and mesa-filesystem-24.1.2-8.fc40.x86_64 from @System
  - conflicting requests
  ```
I support force merging if needed.

**Comment by m2Giles** [comment]
Identified Issues:

When kernel versions match, vmlinuz is not being copied correctly.
Dracut has output errors due to inability to access /dev/kmsg

Downtream images:
- Nvidia failing to install
- Kernel Swap not completely occuring where rpm reports more than one kernel installed
- Several error: -1 transaction failures

**Review by KyleGospo — APPROVED** [review]

**Comment by m0gg** [comment]
I'm still trying to figure out what's causing dracut to be skipped in the kernel-skew case. It's only run when building on rootful podman.

**Comment by m0gg** [comment]
Found the reason for the weird `rpm-ostree install <kernel>` not triggering a proper initramfs rebuild. Rootless buildah (like it was used in the wrapper) uses `fuse-overlayfs` instead of `overlay`. Either rpm-ostree or dracut seem to not like that. It works totally fine when forcing `overlay`:

```sh
#!/bin/sh
set -eux

# get local graphroot at runtime
GRAPH_ROOT="$(podman info -f json | jq -r '.store.graphRoot')"

exec podman run --rm \
  --privileged \
  --net=host \
  --cgroups=disabled \
  --runtime=crun \
  --runtime-flag=cgroup-manager=disabled \
  --security-opt=label=disable \
  --security-opt=seccomp=unconfined \
  -v "$HOME:$HOME" \
  -v "$GRAPH_ROOT":/var/lib/containers/storage \
  -v "$(pwd):"/builder \
  -v /dev/null:/usr/bin/fuse-overlayfs \
  -v /dev/null:/dev/kmsg \
  -w /builder \
  quay.io/buildah/stable:v1 buildah --storage-opt overlay.mount_program="" "$@"
```

For whatever reason, podman will still use the `fuse-overlayfs` mount helper unless access is denied, hence the `-v /dev/null:/usr/bin/fuse-overlayfs`.

Even though i saw @m2Giles picking up the spirit of this PR in #612, i still think using the buildah container is the superior solution to relying on that weirdly crippled buildah of the ubuntu 24.04 runner.

m2Giles:
I'm willing to possibly try this again, but given the fact that the builds were unexpectedly buggy I would like to see this in working outside of main first.

m0gg:
> I'm willing to possibly try this again, but given the fact that the builds were unexpectedly buggy I would like to see this in working outside of main first.

Could you supply some more information about those unexpected bugs?

PR #615, opened 2024-07-27
fix: remove `chsh` from the images

### LINKED ISSUE #598: [Discussion] Remove chsh
Changing your login shell can cause a number of issues on atomic images. The biggest one is that you will break your user account if you rebase to an image that doesn't have the shell.

There also have been issues with users unable and to login after changing their shell from time to time.

The proposal is to remove the chsh binary from main images.

Most modern terminal emulators let you run a custom command instead of your default shell.

There are downstream effects requiring us to change just scripts that expect to determine shell based on $SHELL variable and ones that changed the default shell

castrojo:
Fixes #598

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -172,7 +172,8 @@
         },
         "exclude": {
             "all": [
-                "ffmpeg-free",
+                "chsh", 
+		"ffmpeg-free",
                 "google-noto-sans-cjk-vf-fonts",
                 "libavcodec-free",
                 "libavdevice-free",


m2Giles:
Need to make sure none of the ujust scripts are still using chsh

castrojo:
https://github.com/ublue-os/config/pull/306

castrojo:
Derp, `chsh` is part of utils-linux, it'd need to be split out upstream, config PR should still be merged though.

Issue #553, opened 2024-04-10
rpm-ostree CVE-2024-2905 (world-readable `/etc/[g]shadow[-]`)


travier:
Due to a bug in rpm-ostree, the `/etc/shadow`, `/etc/shadow-`, `/etc/gshadow` and `/etc/gshadow-` files in Fedora CoreOS, Fedora IoT and Fedora Atomic Desktops have the world-readable bit set.

This mainly impacts systems installed starting from the Fedora 39 release. Older systems that have been updated are not impacted.

To fix impacted systems immediately, run the following command as root:
```
$ chmod --verbose 0000 /etc/shadow /etc/gshadow /etc/shadow- /etc/gshadow-
```

For more details, notably the full list of affected versions, see: https://github.com/coreos/rpm-ostree/security/advisories/GHSA-2m76-cwhg-7wv6

This indirectly impact Universal Blue images as well. The latest container images should have the fix (but I've not tested them yet) thus a rebuild and update is all that should be needed for you.

bsherman:
Status message from 10:42 CDT:
The rpm-ostree versions we should expect to contain the fix on respective releases: https://koji.fedoraproject.org/koji/packageinfo?packageID=18142

F38: rpm-ostree-2023.10-4.fc38
F39: rpm-ostree-2024.4-6.fc39
F40: rpm-ostree-2024.4-5.fc40

So, of all our desktop image rebuilds from today, only F39 picked up the fix as F38/F40 packages are still tagged for `-updates-testing`.

This is a moot point for F38 as those users were not vulnerable.
This is a successful mitigation for F39 as the fixed package includes the systemd service to auto-fix permissions.
F40 is still un-mitigated, but users can do so manually.

noelmiller:
> Status message from 10:42 CDT: The rpm-ostree versions we should expect to contain the fix on respective releases: https://koji.fedoraproject.org/koji/packageinfo?packageID=18142
> 
> F38: rpm-ostree-2023.10-4.fc38 F39: rpm-ostree-2024.4-6.fc39 F40: rpm-ostree-2024.4-5.fc40
> 
> So, of all our desktop image rebuilds from today, only F39 picked up the fix as F38/F40 packages are still tagged for `-updates-testing`.
> 
> This is a moot point for F38 as those users were not vulnerable. This is a successful mitigation for F39 as the fixed package includes the systemd service to auto-fix permissions. F40 is still un-mitigated, but users can do so manually.

[...]

PR #631, opened 2024-08-19
fix: Include fuse package for AppImage support

KyleGospo:
https://github.com/fedora-silverblue/issue-tracker/issues/589

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -11,6 +11,7 @@
                 "ffmpeg-libs",
                 "ffmpegthumbnailer",
                 "flatpak-spawn",
+		"fuse",
                 "fzf",
                 "google-noto-sans-balinese-fonts",
                 "google-noto-sans-cjk-fonts",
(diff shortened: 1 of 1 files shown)


p5 (approved):

noelmiller:
Before we merge this, let's just keep an eye on https://gitlab.com/fedora/ostree/sig/-/issues/50 so we don't add this package unnecessarily. I know it will just skip it if it's already there, but if upstream is going to handle it, that makes one less package :)

gallium-stanza:
Merged upstream
[#553 common: Keep fuse for AppImage compatibility](https://pagure.io/workstation-ostree-config/pull-request/553)

packages.json
```diff
@@ -11,6 +11,7 @@
                 "ffmpeg-libs",
                 "ffmpegthumbnailer",
                 "flatpak-spawn",
+		"fuse",
```
travier, line 14: Nit: tabs vs space

p5 (approved):

noelmiller (approved):

PR #635, opened 2024-08-28
fix: Correct rpm-ostree issue that prevents rollbacks

KyleGospo:
Upstream issue: https://github.com/coreos/rpm-ostree/issues/5048
Upstream MR: https://github.com/coreos/rpm-ostree/pull/5069
Package source: https://github.com/ublue-os/packages/tree/main/staging/rpm-ostree

Also sneaks in a quick fix for some whitespace issues.


p5 (approved):
LGTM!

bsherman (commented):
I looked at our rpm-ostree.spec file in https://github.com/ublue-os/packages/blob/main/staging/rpm-ostree/rpm-ostree.spec

It doesn't match upstream https://github.com/coreos/rpm-ostree/blob/main/packaging/rpm-ostree.spec

New requirements: 
- `fuse3` package
- `usr/bin/setpriv`

Dropped requirements:
- `libzstd-devel`

Is this intended?

KyleGospo:
> Is this intended?

Yes, it's copied from Fedora.
https://src.fedoraproject.org/rpms/rpm-ostree/tree/f40

bsherman:
> > Is this intended?
> 
> Yes, it's copied from Fedora. https://src.fedoraproject.org/rpms/rpm-ostree/tree/f40

Yep, I just went down that path, to see if that's where you pulled it. I agree, this spec matches.

Thanks!

bsherman (approved):
This should be good.

PR #638, opened 2024-09-18
feat: enable Fedora 41 builds

p5:
This PR enables the Fedora 41 image builds.  They are still not ready to be used by end-users since some key packages are missing (such as the hardware acceleration ones), but these should be enabled before we switch the tags to latest.

We still need to switch around the LATEST, STABLE and GTS labels to the new images before this is GA, but this will be in a different PR.


m2Giles:
We also need to turn on akmods

p5:
Current missing packages are:
- libheif
- libvdpau
- intel-vaapi-driver
- pipewire-libs-extra

All from the Negativo multimedia repo (https://negativo17.org/repos/multimedia/fedora-41/x86_64/)

We'll merge this PR for now, since we need something to test the changes on.  This image is not yet supported by us nor upstream, so end users shouldn't rebase to these

castrojo (approved):
Let's goooooooooooo!

m2Giles (approved):

PR #636, opened 2024-09-06
feat: use negativo17 in lieu of rpmfusion

bsherman:
This PR replaces our uses of the rpmfusion package repo with negativo17's fedora-multimedia repo.

The two motivations for this:
1) reduce number of upstream repos on which we depend. (We've already moved to negativo17 for nvidia since the nvidia packaging allowed more choice in which components to install, so we are quite committed in that regard)
2) generally avoid "package skew".  Though we have had plenty of skew with rpmfusion, I do not expect that simply changing to a different package provider will solve this. However, there is a chance this PR will help, since much of our "skew" has been around mesa and ffmpeg. This PR changes how we swap out mesa for "less free" versions, and I think it should work more reliably. 

I've run builds locally and inspected the changed packages (eg, libheif, ffmpeg, gstreamer1-plugins, pipewire bluetooth, libva and intel drivers, etc) to make sure the new packages provide what was there before.. Instead of some package names being replaced in `packages.json` they are now "override-replaced" in the `install.sh` . Also, some package name changes are due to  negativo17's conventions or that their packages have more files in a single package vs fedora (eg, libheif from negativo17 has the contents of the both libheif and libheif tools from before).

PR #645, opened 2024-10-06
chore: revert change of gstreamer-plugins

bsherman:
In PR #636 I was over-eager in changing out "-free" packages which had alternatives in negativo17's fedora-multimedia repo.

Specifically, the gstreamer-plugins provided have a large set of dependencies which increase the size of the "*-main" images by approximately 1GiB. Reverting that change as if a user or downstream wishes it, they can make the change (eg, Bazzite does this I think), but others should not be burdened with this size change, expecially not as part of a PR which was just supposed to swap out the source of some packages.


m2Giles (approved):
These should be set in bluefin/bazzite!

EyeCantCU (approved):

PR #648, opened 2024-10-09
feat: include rar from negativo17

qoijjj:
Sadly, rar from brew is only available via cask and so doesn't work on linux: https://formulae.brew.sh/cask/rar

Thankfully, negativo17 provides this package in the multimedia repo, which was just added. The `rar` package is less than half a megabyte, so it shouldn't meaningfully impact image sizes.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -43,6 +43,7 @@
                 "pipewire-libs-extra",
                 "pipewire-plugin-libcamera",
                 "powerstat",
+                "rar",
                 "smartmontools",
                 "solaar-udev",
                 "squashfs-tools",


qoijjj:
@bsherman it is explicitly in scope, moreso than many of the packages already included (like tmux):

```
Non-free codecs aka. “a bunch of that RPMFusion stuff”.
```

This is adding support for a proprietary file format. It is expressly in scope.

If we suddenly want to make this project adhere more closely to the scope, rar should be added and tmux and other packages would need to be removed.

bsherman:
> ```
> Non-free codecs aka. “a bunch of that RPMFusion stuff”.
> ```
> 
> This is adding support for a proprietary file format. It is expressly in scope.

This is a reasonable view of the scope. Thank you, I want to have that clear for anything we consider adding to main.

> If we suddenly want to make this project adhere more closely to the scope, rar should be added and tmux and other packages would need to be removed.

It has been argued several times that the package list should either be made "more compliant with scope" else we should be more permissive about adding packages because some existing package doesn't fit well with scope.

My position has been consistent, nothing "sudden" about it:
- existing packages have grandfathered status even if they don't fit clearly in scope
- existing packages are always subject to removal over time if deemed appropriate
- new package requests will always be weighed against scope

We will not be doing a full correction of existing image packages to comply with one's perspective of appropriate scope. 
However, we HAVE pruned things out of the package list as seems reasonable... hence being "brutal about making things out of scope."

qoijjj:
@bsherman makes sense, yeah I was just using it as an example. I don't really care about tmux being present.

PR #649, opened 2024-10-09
fix: bug in negativo prioritization command

qoijjj:
as-is, this changes the srpm repo's priority instead of the rpm repo:

```
[fedora-multimedia-source]
name=negativo17 - Multimedia - Source
baseurl=https://negativo17.org/repos/multimedia/fedora-$releasever/SRPMS
enabled=1
priority=90
```


bsherman (approved):
Thank you for catching this one!

castrojo (approved):

PR #660, opened 2024-10-30
fix(kinoite): Include missing KDE image format codecs

NiHaiden:
This includes the missing JPEG-XL and other codec support which is present in F40 Kinoite Builds but not in the F41 builds. Downstream Issue Example: https://github.com/ublue-os/bluefin/issues/1811

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -133,7 +133,10 @@
     "41": {
         "include": {
             "all": [],
-            "kinoite": []
+            "kinoite": [
+                "kf6-kimageformats",
+                "qt6-qtimageformats"
+            ]
         },
         "exclude": {
             "all": []


packages.json
```diff
@@ -133,7 +133,10 @@
     "41": {
```
fiftydinar, line 133: This can be removed in `40` now, since it will be included in all kinoite versions:

```json
    "40": {
        "include": {
            "all": [],
            "kinoite": []
```

NiHaiden: But wouldn't that break the support on current Fedora 40 based images?

fiftydinar: I made a mistake in looking, sorry

fiftydinar (approved):
This is good to go

noelmiller (approved):

bsherman:
> This includes the missing JPEG-XL and other codec support which is present in F40 Kinoite Builds but not in the F41 builds. Downstream Issue Example: [ublue-os/bluefin#1811](https://github.com/ublue-os/bluefin/issues/1811)

For additions to main, I am pretty uptight about always calling out how this fits in our project/image scope.

In this case, I do think @NiHaiden provided the information in the description.

This isn't a net-add, rather it's maintaining expected behavior from F40 in F41 builds.

I'm good with this.

:+1:

nisel11:
kf6-kimageformats have [some problems](https://github.com/ublue-os/aurora/issues/8#issuecomment-2549680711)

PR #661, opened 2024-10-30
chore: Remove `avif` thumbnailer, as `heif` thumbnailer caches it alr…

fiftydinar:
…eady

Screenshot From 2024-10-30 18-37-29

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -76,6 +76,9 @@ chmod +x ${CSFG}
 # would duplicate warnings provided by ublue already. we don't want confusion
 rm -f /usr/libexec/gnome-software-dkms-helper
 
+# Remove avif thumbnailer, as heif thumbnailer already caches it
+rm -f /usr/share/thumbnailers/avif.thumbnailer
+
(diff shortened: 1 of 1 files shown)


KyleGospo (approved):

install.sh
```diff
@@ -76,6 +76,9 @@ chmod +x ${CSFG}
 # would duplicate warnings provided by ublue already. we don't want confusion
 rm -f /usr/libexec/gnome-software-dkms-helper
 
+# Remove avif thumbnailer, as heif thumbnailer already caches it
```
bsherman, line 79: Is there a negative to the current state?

Do both thumbnailers run, thus wasting resources?
If not...
Does one always get chosen instead of the other, or is it random?

fiftydinar: Nothing breaking, thumbnailing still works.

I'm not sure if both thumbnailers are running at the same time.

I think that one gets chosen over the other, but on which terms, I don't know.

bsherman:
I'm going to close this.

I consider it bad practice to remove a file managed by an RPM package. It is work considering if actually required to fix some broken behavior, but this doesn't seem to be required.

PR #663, opened 2024-10-31
fix: Override copr plugin config to use fedora chroot by default

Zeglius:
This override is to handle the default behavior of dnf5 using ID at /etc/os-release to select which chroot gets used to fetch the copr repo.

An example of the behavior displayed without this override in Bazzite:
```
sudo dnf5 copr enable msmafra/hyprland
 https://copr.fedorainfracloud.org/api_3/rpmrepo/msmafra/hyprland/bazzite-41/                        100% | 726.0   B/s | 457.0   B |  00m01s
Chroot not found in the given Copr project (bazzite-41-x86_64).                                                                              You can choose one of the available chroots explicitly:
 fedora-40-aarch64
 fedora-40-x86_64
 fedora-41-aarch64
 fedora-41-x86_64
 fedora-rawhide-aarch64
 fedora-rawhide-x86_64
```

### DIFF
--- a/sys_files/usr/share/dnf/plugins/copr.vendor.conf
+++ b/sys_files/usr/share/dnf/plugins/copr.vendor.conf
@@ -1,0 +1,20 @@
+# This override is to handle the default behavior of dnf5 using ID at /etc/os-release
+# to select which chroot gets used to fetch the copr repo.
+#
+# An example of the behavior displayed without this override in Bazzite:
+# sudo dnf5 copr enable msmafra/hyprland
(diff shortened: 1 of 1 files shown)


fiftydinar (approved):

bsherman:
@Zeglius this PR was accidentally force-merged, thus I've reverted it in #664 .

The fix itself is appreciated, so please re-submit.

I can tell you in advance, the one change I'd request is to use an example COPR repo from the ublue-os org itself rather than a 3rd party.

Thank you.

PR #665, opened 2024-10-31
fix: Override copr plugin config to use fedora chroot by default (resubmission)

Zeglius:
Resubmission of #663 as per #664


KyleGospo (approved):

KyleGospo:
This looks like a good change to me.

It doesn't affect existing users in any way and for images downstream of Main that customize the OS release it prevents DNF from breaking.

bsherman (approved):
Thank you!

PR #670, opened 2024-11-01
chore: proposal refactor

p5:
Just an idea of a refactor to make it slightly easier to work in this repo.

TODO:
- Figure out if I broke the alternatives hacks
- Ensure package script still works after small refactor
- Confirm removal of cosign binary is okay
- Can we just always run initramfs regeneration, no matter the kernel?


Containerfile
```diff
@@ -19,21 +19,16 @@ ARG IMAGE_NAME="${IMAGE_NAME:-silverblue}"
 ARG FEDORA_MAJOR_VERSION="${FEDORA_MAJOR_VERSION:-40}"
 ARG KERNEL_VERSION="${KERNEL_VERSION:-6.9.7-200.fc40.x86_64}"
 
-COPY sys_files/usr /usr
+COPY system_files/usr /usr
 
 RUN --mount=type=cache,dst=/var/cache/rpm-ostree \
-    --mount=type=bind,from=ctx,src=/,dst=/ctx \
+    --mount=type=bind,rw=true,from=ctx,src=/,dst=/ctx \
```
bsherman, line 25: why does is this bind mount changed to `rw=true`?

p5: This was added because I got annoyed with permissions on the scripts/ directory.  The Containerfile now `chmod +x`s all scripts in the directory, which required write access.   It can be removed if we wish to proceed with this.

m2Giles:
If you aren't reliant on exit for control flow, you could source the
scripts instead

Containerfile
bsherman: I like the bind being read-only, seems a bit odd to mutate the scripts on the fly.

Alternatives are:
1. chmod +x the scripts locally and commit the change 
2. `bash /ctx/build_files/scriptname.sh` at runtime instead of `/ctx/build_files/scriptname.sh`

xynydev: Option 2 seems solid, if all the scripts are standardized as bash scripts.

m2Giles: I find it unlikely for us to move off of bash any time soon so #2 would be reasonable for me.

bsherman: I don't really understand why the permissions are problematic in the first place. They are stored in git repo.

Only potential problem is someone committing in windows (not WSL).

bsherman:
I don't really want to see changes in main right now.

It's actually quite simple compared to Bluefin or Bazzite... And we've been working on priorities for the project which include further simplification and streamlining of workflows. 

I would like to pause on refactoring of main until we've made headway on those other priorities.

bsherman:
@p5 that didn't require a close. But, keep the branch around... after we've hit the other priorities maybe it'll be even easier to cleanup.

PR #671, opened 2024-11-02
feat: Replace gnome-software with non-dkms version on F41 and above.

KyleGospo:
Fedora 47 includes a version of gnome-software with a DKMS plugin compiled in. This plugin tries to aid the user by helping setup a MOK for custom kmods, etc... Since Universal Blue has another solution for this which does not include users self-compiling kmods, we needed to disable this.

See: https://copr.fedorainfracloud.org/coprs/ublue-os/staging/package/gnome-software/

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -65,7 +65,9 @@ if [[ "$FEDORA_MAJOR_VERSION" -ge "41" && "$IMAGE_NAME" == "silverblue" ]]; then
     rpm-ostree override remove \
         gnome-software-rpm-ostree
     rpm-ostree override replace \
-        https://download.copr.fedorainfracloud.org/results/ublue-os/staging/fedora-41-x86_64/08203620-gnome-software/gnome-software-47.1-100.ublue.fc41.x86_64.rpm
+        --experimental \
+        --from repo=copr:copr.fedorainfracloud.org:ublue-os:staging \
+        gnome-software
 fi
 
 # run common packages script
--- a/install.sh
+++ b/install.sh
@@ -62,6 +62,8 @@ fi
 
 # Disable DKMS support in gnome-software
 if [[ "$FEDORA_MAJOR_VERSION" -ge "41" && "$IMAGE_NAME" == "silverblue" ]]; then
+    rpm-ostree override remove \
+        gnome-software-rpm-ostree
     rpm-ostree override replace \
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -60,6 +60,12 @@ if [[ "$FEDORA_MAJOR_VERSION" -ne "41" ]]; then
         libvdpau
 fi
 
+# Disable DKMS support in gnome-software
+if [[ "$FEDORA_MAJOR_VERSION" -ge "41" && "$IMAGE_NAME" == "silverblue" ]]; then
+    rpm-ostree override replace \
+        https://download.copr.fedorainfracloud.org/results/ublue-os/staging/fedora-41-x86_64/08203620-gnome-software/gnome-software-47.1-100.ublue.fc41.x86_64.rpm
```
bsherman, line 66: This is a brittle way to override, no?

Shouldn't we override by override replace from the ublue-os-stabing COPR?

KyleGospo: This was a test and it failed, you can ignore it. We need a way to get around it crashing while trying to override.

bsherman: I think I see that the problem was doing it with a replace that pulls from the repo ends up pulling in a bunch of devel packages, no?

That seems like it could be related to rpm-ostree always pulling recommended packages. I'm not sure.

Well, we can probably merge as is, as fairly soon, I expect we'll be doing an overhaul of main builds to use dnf, and I hope this will be easier to address at that time.

bsherman (approved):

install.sh
KyleGospo: This fails to build as is, so it needs some kind of fix. Marked as draft

bsherman (changes requested):
Oops... i see this doesn't work... we need to figure this out. :-/

install.sh
bsherman: I'll test.

fiftydinar: It works now.
It's interesting seeing how `gnome-software-rpm-ostree` is getting removed, then getting installed with Ublue's version, then gets removed again lol

bsherman: Yes, thank you for the tip, i'm going to see if we can use the repo override now, rather than the direct URL to package.

Yeah, not the only place we remove stuff to have it re-added and removed again.

bsherman: Fixed!

bsherman (approved):

noelmiller (approved):

bsherman:
Testing a change before we merge this.

PR #679, opened 2024-11-21
chore: remove cliwrap and call dracut directly

p5:
Removes cliwrap.

I have tested this on Fedora 41 only, and it works as expected.


tulilirockz:
Oh wow I thought this was gonna be much harder

fiftydinar (approved):
Works well

tulilirockz:
From a comment by @bsherman on discord, we probably shouldn't do this yet since it affects everyone. We probably should test it out on other images, and if we notice theres no meaningful difference, then merge it

p5:
I'm going to get this PR into a good state, then we can merge when we feel like it

KyleGospo (approved):

castrojo (approved):

PR #687, opened 2024-12-02
chore: switch to official upstream images

### LINKED ISSUE #669: use official upstream Atomic Desktop images (once they support bootc) insted of current unofficial
This is pending Fedora F41 turning on bootupd for BIOS systems

Also these changes may be related:
https://pagure.io/fedora-infra/ansible/pull-request/2337
https://pagure.io/releng/issue/12142

p5:
Closes #669 

This is fully backwards-compatible.  Builds official images using the official Quay repository, but keeps building the unofficial images using the unofficial repository.

I have tested this with a Fedora 41 image, which worked exactly the same as before.  I haven't tested with Fedora 40.


.github/workflows/reusable-build.yml
```diff
@@ -34,6 +34,15 @@ jobs:
...
           - vauxite
+        include:
+          - image_name: silverblue
+            base_image: quay.io/fedora/fedora-silverblue
+          - image_name: kinoite
+            base_image: quay.io/fedora/fedora-kinoite
+          - image_name: sericea
+            base_image: quay.io/fedora/fedora-sway-atomic
+          - image_name: onyx
+            base_image: quay.io/fedora/fedora-budgie-atomic
```
castrojo, line 45: Should we just go with the upstream names so they end up being `sway-main` and `budgie-main`?

castrojo:
I see `containers.bootc = 1` on the F40 images as well, did they push those as well? Sorry I haven't been keeping track of this as well as I should have!

.github/workflows/reusable-build.yml
p5: We can do, though it may be best waiting for the next major version rather than introducing breaking changes in the middle of a cycle

tulilirockz:
What about `base` being fedora-bootc? is that backward compatible or we cant use it yet?

p5:
> What about `base` being fedora-bootc? is that backward compatible or we cant use it yet?

fedora-bootc is extremely minimal.  I'd probably say too minimal for us at this time.  
Current `base` contains all [these packages](https://gitlab.com/fedora/ostree/ci-test/-/blob/main/common-packages.yaml?ref_type=heads), whereas bootc-base (the tier-0 and tier-x images) contain [far less](https://gitlab.com/fedora/bootc/base-images/-/tree/main/tier-0?ref_type=heads)

fiftydinar:
But the official quay image contains `bootc` also. So idk what's the differentiation for the `fedora-bootc` image to be called like that.

.github/workflows/reusable-build.yml
```diff
@@ -43,12 +58,17 @@ jobs:
...
         shell: bash
         run: |
-          if [[ "${{ matrix.image_name }}" == "mate" ]]; then
-              echo "SOURCE_IMAGE=base" >> $GITHUB_ENV
+          // Overwrite the SOURCE_ORG if set in the matrix.
+          if [[ "${{ matrix.source_org }}" == "fedora-ostree-desktops" ]]; then
+            echo "SOURCE_ORG=${{ matrix.source_org }}" >> $GITHUB_ENV
+          fi
+
+          // Overwrite the SOURCE_IMAGE to prepend fedora- (the new naming convention)
```
p5, line 66:
```suggestion
          if [[ "${{ matrix.source_org }}" == "fedora-ostree-desktops" ]]; then
            echo "SOURCE_ORG=${{ matrix.source_org }}" >> $GITHUB_ENV
          fi

```

PR #718, opened 2025-02-18
chore: remove F39 workflow and refactor matrixes

p5:
This removes the F39 workflow since it's disabled now 39 is EOL
Also refactors the matrix out of the reusable build, which makes it far easier when we come to deprecating the unofficial desktops.

Once merged, I'll make a F42 draft PR where we actually stop building for these DEs in F42+.


.github/workflows/build-40.yml
```diff
@@ -1,15 +1,32 @@
...
+      matrix:
+        image_name:
+          - base
+          - silverblue
+          - kinoite
+          - sericea
+          - onyx
+          - lazurite
+          - vauxite
     with:
```
p5, line 30: Once F42 is out, we can simply duplicate this workflow and remove the images we don't wish to build

.github/workflows/build-41.yml
```diff
@@ -1,15 +1,44 @@
...
+          - image_name: base
+            source_image_name: base-atomic
+          - image_name: onyx
+            source_image_name: budgie-atomic
+          - image_name: sericea
+            source_image_name: sway-atomic
+          - image_name: lazurite
+            source_image_name: lxqt-atomic
+          - image_name: vauxite
+            source_image_name: xfce-atomic
```
p5, line 40: From F42+, we can switch to the *-atomic naming wherever used upstream, so this will only be used in F41 workflows.

.github/workflows/reusable-build.yml
```diff
@@ -21,31 +26,6 @@ jobs:
...
-          - image_name: base
-            source_image: ${{ inputs.fedora_version == '41' && 'base-atomic' || '' }}
-          - image_name: sericea
-            source_image: ${{ inputs.fedora_version == '41' && 'sway-atomic' || '' }}
-          - image_name: onyx
-            source_image: ${{ inputs.fedora_version == '41' && 'budgie-atomic' || '' }}
-          - image_name: lazurite
-            source_image: ${{ inputs.fedora_version == '41' && 'lxqt-atomic' || '' }}
-          - image_name: vauxite
-            source_image: ${{ inputs.fedora_version == '41' && 'xfce-atomic' || '' }}
```
p5, line 47: All matrix logic is in the parent workflows so you don't need any conditionals here

.github/workflows/reusable-build.yml
```diff
@@ -54,19 +34,17 @@ jobs:
...
-              SOURCE_IMAGE=$MATRIX_SOURCE_IMAGE
+          SOURCE_IMAGE=$IMAGE_NAME
+          if [[ "$SOURCE_IMAGE_NAME" != "" ]]; then
+            SOURCE_IMAGE=$SOURCE_IMAGE_NAME
           fi
 
           echo "SOURCE_ORG=fedora-ostree-desktops" >> $GITHUB_ENV
-          echo "IMAGE_NAME=$MATRIX_IMAGE_NAME-main" >> $GITHUB_ENV
+          echo "IMAGE_NAME=$IMAGE_NAME-main" >> $GITHUB_ENV
           echo "SOURCE_IMAGE=$SOURCE_IMAGE" >> $GITHUB_ENV
```
p5, line 48: This can be removed when F41 is EOL'd.  It's needed so we can keep the old "sericea" etc  naming convention for the existing images

castrojo (approved):

bsherman (approved):

Issue #720, opened 2025-02-19
Deprecate unofficial spins from F42+


p5:
A proposal initially made in #669.

We're currently building every image variant made available to us in the quay.io/fedora-ostree-desktops registry, which includes variants with only a handful of users (i.e. XFCE, LXQt etc).  

The proposal is to reduce the number of spins we build against to only those officially supported by Fedora.  These being:
- Silverblue
- Kinoite
- Sway Atomic
- Budgie Atomic

We will also provide a DE-less image for those who wish to use the other desktop environments, where they can `dnf install` packages they require.  This approach has been used for Bluefin LTS, and is proven to work well.

If agreed, this proposal will be aimed at Fedora 42+ images - we will continue to build the XFCE/LXQT/etc images for Fedora 40 and Fedora 41 until they turn End Of Life.

---

This change will also switch to following the upstream naming convention for the images.  This removes the "onyx" and "sericea" naming convention, switching to "budgie-atomic" and "sway-atomic".

castrojo:
+1 let's do this!

bsherman:
+1  I think this was discussed a bit on a planning call back in September/October, but we never made an official proposal. I support the proposal as we've been consistently working to reduce maintenance burdens and improve the quality of what we do provide.

ledif:
+1 from me.

As a side note, I maintain my own Xfce atomic image that I daily drive alongside Aurora: https://github.com/winblues/vauxite. This is just to note for anyone that stumbles across this issue and is looking for an Xfce ublue alternative.

castrojo:
This is done now afaict?

m2Giles:
41 still has lazurite and vauxite. They will go away with 41.

We also have moved to the new naming scheme on 42.

PR #722, opened 2025-02-21
chore: enable F42 images

p5:
Part of #720 

So we can finalise the images to build, and the naming of each.  I don't intend on merging this yet.
The F42 workflow is disabled since we do not expect it to be buildable any time soon.

One thing I'm not sure about is what do we want to call the base image?  base-atomic, or just base?


m2Giles:
I'm personally a fan on leaving it as base.

m2Giles:
With akmods merged, is this now ready?

p5:
> With akmods merged, is this now ready?

Yes it is.  Currently with family, but will finish this in an hour or so.

Just need to enable the workflow triggers and see what happens.

fiftydinar:
> With akmods merged, is this now ready?

Outside of the PR, we need to wait for all used negativo packages to build for F42

castrojo:
Should we just dnf5 this thing?

m2Giles:
I vote yes

fiftydinar:
> Should we just dnf5 this thing?

Posted in Discord, I'll also post here.

I grab `main` manually instead of using a base image & here's how I utilize `dnf5` in it:

https://github.com/fiftydinar/gidro-os/blob/baf227a04a2da83d3dd6053f87fc7c0e2af6d008/recipes/base.yml

replace = `distro-sync`
replace old/new = `swap` 

I don't grab all of the `main`, but it should be a good reference.

castrojo (approved):

KyleGospo (approved):

PR #746, opened 2025-02-23
chore: rm packages copr /cp containers/policy.json

m2Giles:


install.sh
```diff
@@ -28,6 +28,8 @@ rpm-ostree install \
     /tmp/akmods-rpms/*.rpm \
     fedora-repos-archive
 
+cp /usr/etc/containers/policy.json /etc/containers/policy.json
```
p5, line 31: Can we make this a `mv`?
We should delete `/use/etc` entirely so it satisfies `bootc container lint`

bsherman (approved):

p5 (approved):

install.sh
m2Giles: Made it a move, do we need to completely nuke the /usr/etc tree?

p5: I'm not sure.  I didn't think empty directories were kept in OCI images, but I may be mistaken

bsherman: That does leave our RPM (ublue-os-signing` in a weird place where it owns a file in `/usr/etc/containers/policy.json` which will never exist.

Should we instead solve this in the RPM?

Maybe it owns the specific file we want but in a new name like `/etc/containers/policy-ublue.json` and in the `%post-install` section it does this copy... then anyone installing this "just works" though it does clobber the stock file.

m2Giles: Yeah, I feel like clobbering it like that could be an unexpected side effect. I think rpm can handle missing files since we are explicitly doing that in akmods with the kernel signing and kmod signing.

bsherman: While I'd still like to see the RPM improved, I'm fine with this for now. It's still a clear improvement.

bsherman (approved):

PR #755, opened 2025-03-07
chore(packages): Remove `gnome-epub-thumbnailer`

fiftydinar:
It's already included by default.


castrojo (approved):

fiftydinar:
It's installed by default in F41, but not in F40.

Same goes for `kernel-tools`, which we removed recently.

bsherman:
> It's installed by default in F41, but not in F40.

So sounds like we should remove this after we drop the last F40 build, as then it will be a net zero change.

p5 (approved):

PR #767, opened 2025-03-22
chore: cleanup packages.sh

m2Giles:
packages.sh and packages.json are designed around the behavior of rpm-ostree. Namely, it can install multiple packages while removing multiple packages in a single transaction. `dnf5` does not support this. `dnf5 distro-sync` can bring multiple packages into sync at once, but cannot be used to install packages. `dnf5 swap` only works on one package at a time.

This cleanup removes the behaviour of doing a single transaction of installing and removing. Instead, it parses the the json file to install packages first. Then, it parses the json file and confirms the packages are on image before removing them.

Already merged is a basic package checker for some of the major packages for each image (F42 images only. Lazurite and Vauxite are on their own right now.)

This is a short term change right now that slightly changes functionality. We already are installing items in install.sh seperate from this json being the ultimate source of truth.

### DIFF
--- a/build_files/packages.sh
+++ b/build_files/packages.sh
@@ -1,53 +1,34 @@
-#!/bin/sh
+#!/bin/bash
 
 set -ouex pipefail
 
-RELEASE="$(rpm -E %fedora)"
-
 # build list of all packages requested for inclusion
-INCLUDED_PACKAGES=($(jq -r "[(.all.include | (.all, select(.\"$IMAGE_NAME\" != null).\"$IMAGE_NAME\")[]), \
+readarray -t INCLUDED_PACKAGES < <(jq -r "[(.all.include | (.all, select(.\"$IMAGE_NAME\" != null).\"$IMAGE_NAME\")[]), \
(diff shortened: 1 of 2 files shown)


castrojo (approved):

bsherman (approved):
One comment, the old code could remove packages before installing new and then double check it was removed again and not added as a dep. 

Maybe not needed with dnf but I know we used that, if not in bluefin, somewhere else. 

This is good though. We should improve with this and see if we need to refine further.

m2Giles:
> One comment, the old code could remove packages before installing new and then double check it was removed again and not added as a dep. 
> 
> Maybe not needed with dnf but I know we used that, if not in bluefin, somewhere else. 
> 
> This is good though. We should improve with this and see if we need to refine further. 

Unfortunately, that was an item that we explicitly removed due to dnf5 not having a similar multiple packages replace like rpm-ostree.

With rpm-ostree we could swap several packages as part of one transaction (that had arbitrary names). Dnf5 on the other hand has the limitation of only swapping one package at a time if the names differ. It will swap packages if they have the same name and why we end up having ffmpeg in the install block and not as a separate swap in install.sh like the mesa ones.

But there is definitely room for improvement. I would like to get package swaps out of install.sh eventually.

PR #770, opened 2025-03-23
feat: cleanup justfile and use just to build

m2Giles:
Gets this working again. Does not have an ISO builder yet as we are between several options right now. This also has most of the stuff necessary to convert the github workflow to use the justfile for building images.


Justfile
```diff
@@ -1,49 +1,577 @@
...
+    elif [[ "$fedora_version" -eq "{{ beta }}" ]]; then
+        COMMIT_TAGS+=("$SHA_SHORT-beta")
+        BUILD_TAGS=("beta")
+    fi
+
+    # Get Fedora CoreOS Version
+    stable_version="$(skopeo inspect docker://quay.io/fedora/fedora-coreos:stable | jq -r '.Labels["ostree.linux"]' | grep -oP 'fc\K[0-9]+')"
+    if [[ "$fedora_version" -eq "$stable_version" ]]; then
+        BUILD_TAGS+=("stable")
+    fi
```
p5, line 248: I don't think we should tag with coreos tags here.  CoreOS tags mean we install the CoreOS kernel version, which isn't what we do in main

m2Giles: Removed.

m2Giles:
This has the scaffolding for a complete switchover to using just for build actions.

Additionally, re-adds devcontainer support.

sbom items are templated, but not enabled due to space constraints on the runner.

m2Giles:
I think the list-tags loop is a bit noisy and will write it to a file instead.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,43 @@
...
+				"mads-hartmann.bash-ide-vscode",
+				"ms-azuretools.vscode-docker",
+				"nefrob.vscode-just-syntax",
+				"timonwong.shellcheck"
+			]
+		}
+	},
+	"runArgs": [
+		// Only Needed for Podman in VSCode (DevPod autohandles)
+		// "--userns=keep-id",
```
bsherman, line 37: This is a side question, not really a comment on the PR.

If `--userns=keep-id` is added as is needed for Podman in VSCode, does that break DevPod (as DevPod works fine without this)?

m2Giles: It is needed if you are using podman as your docker provider for devcontainers in VSCode.

--userns=keep-id is an incompatible argument for docker.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,43 @@
+// For format details, see https://aka.ms/devcontainer.json.
+{
+	"$schema": "https://raw.githubusercontent.com/devcontainers/spec/refs/heads/main/schemas/devContainer.schema.json",
+	"name": "m2os-devcontainer",
```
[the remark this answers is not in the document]
bsherman: I propose `ubluemain-devcontainer`

.devcontainer/devcontainer.json
bsherman: Ah I see, maybe we clarify that comment a bit more?

I am pretty interested in using devcontainers more... but I haven't yet dug in very deep. Thanks for these examples!

Justfile
```diff
@@ -1,49 +1,581 @@
...
+    ["main"]="main"
+    ["nvidia"]="nvidia"
+)'
+
+# Sudo/Podman/Just
+
+[private]
+SUDO_DISPLAY := env("DISPLAY", "") || env("WAYLAND_DISPLAY", "")
+[private]
+SUDOIF := if `id -u` == "0" { "" } else if SUDO_DISPLAY != "" { which("sudo") + " --askpass" } else { which("sudo") }
```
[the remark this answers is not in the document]
m2Giles: Installing syft/cosign if it's not installed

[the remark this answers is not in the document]
m2Giles: There is the thought that maybe this should be removed and instead require you to install the tools if they are not present.

.devcontainer/devcontainer.json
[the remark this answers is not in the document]
m2Giles: I changed it to ublue-main-devcontainer.

I think we may want to centralize these because we could reuse a prebuilt image instead.

bsherman (approved):
I think this looks solid. 

With any significant refactor to the build like this, there's a risk of unexpected changes, but I'm very happy with the direction this takes us. Both the use of just and devcontainer are great patterns which aid in local build development and testing.

p5 (approved):

m2Giles:
My goal is to bring the devcontainer pattern to the other repos.

The main thing I would like to do is centralize it so that instead it can pull the image instead of duplicating the config in each repo.

If podman machine gets working, then a rootless podman instance could also do ISO building.

PR #772, opened 2025-03-25
feat: add git-email

tidux:
This adds the Git email utilities so `git send-email` works.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -10,6 +10,7 @@
                 "flatpak-spawn",
                 "fuse",
                 "fzf",
+                "git-email",
                 "heif-pixbuf-loader",
                 "htop",
                 "just",
--- a/packages.json
+++ b/packages.json
@@ -1,165 +1,143 @@
 {
-  "all": {
-    "include": {
-      "all": [
-        "alsa-firmware",
-        "android-udev-rules",
-        "apr",
-        "apr-util",
-        "distrobox",
-        "flatpak-spawn",
-        "fuse",
-        "fzf",
-        "git-email",
-        "heif-pixbuf-loader",
-        "htop",
-        "just",
-        "libcamera",
-        "libcamera-tools",
-        "libcamera-ipa",
-        "libfdk-aac",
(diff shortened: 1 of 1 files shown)


tulilirockz (approved):
Honestly, sounds good to me, seems like a particularly small package and its handy for doing stuff on the host. Although we can maybe use homebrew for this?

Seems a tiny bit redundant but id be up to having this on the host

tidux:
> Honestly, sounds good to me, seems like a particularly small package and its handy for doing stuff on the host. Although we can maybe use homebrew for this?

I looked at that and there's no Homebrew package for git-email.  The MacOS instructions for configuring `git send-email` involve invoking a CPAN command with system `perl`.

bsherman:
I don't have a strong feeling for or against this package, but we've been _incredibly_ strict about adding any packages to main for a long time. I don't think we've added one in over a year except as a replacement/bug fix, but we have removed things. Essentially, we've called `main` images "done" and added things only to downstreams (eg, Aurora, Bazzite, Bluefin, uCore) unless there's been a strong consensus.

Given that, I think we need a more significant number of approvals than the typical 2 in order to add this.

@ublue-os/approver  Thoughts?

bsherman (changes requested):
I'm requesting a few more than two approvals before we merge this due to our "main is done" stance.

antheas:
Would be good for dx images, but bloat for the rest

GarciaLnk:
-1 IMHO, `brew install git` already provides this and there's no benefit to having it baked in.

Screenshot_20250326_010746

tidux:
I raised this because I didn't see a git-email package in Homebrew.  If `brew install git` provides the functionality I'll withdraw the PR.  Maybe add a hint to the Bluefin dx docs like there is for NeoVim?

m2Giles:
I'm in the no camp.

Upstream already only uses a significant subset of git by default. This will pull in all of git's perl dependencies.

For a user that needs this functionality. Brew is available. A distrobox could also be used.

PR #776, opened 2025-03-29
chore: add intel hw-accel to f42

m2Giles:
These are all in multimedia F42.


fiftydinar (approved):

p5 (approved):

bsherman (approved):
Got it. We don't need to add/replace these in F42, but do need to continue doing so in F41.

Looks good.

PR #782, opened 2025-04-06
fix(packages): remove gnome-tweaks

castrojo:
Refine (flatpak) does a better job than this, it's time to move to it.


bsherman (dismissed):
I think we can merge this once both Bazzite and Bluefin have Refine added their defaults... but probably we should do this as part of the F42 release, so coming fast!

p5 (approved):

tulilirockz (approved):

PR #788, opened 2025-04-16
feat: add cosmic build

bigpod98:
added cosmic-atomic to build-42.yaml


p5 (approved):
Happy to have cosmic main images.

ledif (approved):
Since Cosmic is now an official Fedora Atomic variant on https://fedoraproject.org/atomic-desktops, it makes sense to include it in the matrix now.

Issue #691, opened 2024-12-14
[Proposal] Consolidate Main, Config, HWE, Akmods, and Kernel Cache


m2Giles:
As Aurora/Bluefin/Bazzite/Ucore developed, we have dropped usage of hwe internally. Nvidia/asus/surface/hwe images are generated in downstream images using the scripts directly instead of using HWE as a caching layer. Asus and surface images have not been building for F41.

Proposal: Drop HWE repo and just build Nvidia images in main. Do not build images for Asus/Surface.

Config is currently it's own repo. It effectively generates 6 RPMs for consumption and one of them should actually be a file copy. Building RPMs is the correct way to include additional items for our images; however, we have sprawled quite a bit on what we do inside of config.

Proposal: Drop Config repo and include in main.

Akmods needs to be tied to kernel-cache. We have already agreed to merged them but have yet to implement.

Proposal: Continue with Merge and consolidate into main.

This will mean that effectively all of ublue that we and others consume would live in this one Repo.

Negatives: Issues board will be have more traffic. This could be resolved with tagging.

bsherman:
Updated thoughts on this proposal.

Current state of things:
- `hwe` has been announced deprecated as of F42. We struggled to maintain it, so it's going away. `nvidia` image builds (and the `nvidia-install.sh` script) will get moved to `main`
- `kernel-cache` and `akmods` have already been merged into combined `akmods` repo and images (see https://github.com/ublue-os/akmods/pull/295)
- no significant changes have been made to `config` related to this proposal


My suggestions are:
1. Take the next step to cleanup `config` and `main` by merging them. 
2. The `config` RPM artifacts should build on the same fedora release as the main image (currently they are all building on older which seems off)
3. After merging `config` and `main`, move `*-nvidia` image builds and `nvidia-install.sh` to `main` repo.
4. Archive HWE with F42 release as currently planned.
5. Hold off on any further `akmods` / `main` merge consideration until this set of changes has settled out.

castrojo:
Updates to this, ublue-os/config has been merged into ublue-os/packages. This is to centralize it since we're generating RPMs out of all the rules.

m2Giles:
Yes, we ended up going a slightly different direction with Config than a straight merge into main; however, it is still consolidation and centralization.

brontium:
[...]

PR #813, opened 2025-04-25
chore(nvidia): moving nvidia builds from hwe to main

bsherman:
This adds nvidia-install.sh to the main repo and configures workflows to build and publish `ublue-os/*-nvidia` images in this repo as well.

- adds `nvidia-install.sh` from `hwe` repo
- adds an `image_variant` to the build matrix to define build of `main` or `nvidia` image
- uses `image_variant` to modify built image name and pass a flag to buildah to conditionally run `nvidia-install.sh`
- adds a `FROM akmods-nvidia-open` in `Containerfile` which uses the new renovate/digest pattern
- `akmods-nvidia-open`  bind mounts into main build but only used when BUILD_NVIDIA condition is set 

Relates: #691


bsherman:
Notes: before merging this, we should stop all nvidia builds on HWE repos to avoid collisions.

Also, Aurora, Bluefin, Bazzite all need PRs to switch to this repo as source for `nvidia-install.sh`, but since I'm not removing it from `hwe`, that's not a blocking issue.

p5 (dismissed):

p5:
> Also, Aurora, Bluefin, Bazzite all need PRs to switch to this repo as source for nvidia-install.sh, but since I'm not removing it from hwe, that's not a blocking issue.

Would we do that, or would we FROM the images in this repo?  I can't remember why we switched to calling the script in each image

bsherman:
> > Also, Aurora, Bluefin, Bazzite all need PRs to switch to this repo as source for nvidia-install.sh, but since I'm not removing it from hwe, that's not a blocking issue.
> 
> Would we do that, or would we FROM the images in this repo? I can't remember why we switched to calling the script in each image

The downstream products call the script because it's more efficient and was less prone to failure, IIRC.

> I guess if we reference the script from this repo, my next question would be why do we build the images in the first place?

You ask a good question. I've actually started wondering why we build main images, much less nvidia. At the moment, my answer is precedent. We've been building these, so we should keep doing so until we decide against it. Since we haven't proposed to not build them, and we specifically said we would in the related issue, that's what I did.

bsherman:
> Notes: before merging this, we should stop all nvidia builds on HWE repos to avoid collisions.

Correction to this comment:  I think we should stop ALL builds on HWE repo because the only thing building there now are nvidia builds. @m2Giles please correct me if I'm wrong.

Also, once I've PR'd the new repo change for `nvidia-install.sh` into all downstreams, we can then archive HWE.

m2Giles:
HWE should halt. This picks up the Nvidia builds.

bsherman:
> HWE should halt. This picks up the Nvidia builds.

Yep, I was just clarifying that there's only nvidia building in HWE at this time.

bsherman:
I believe all the `*-nvidia` container package settings have now been updated to allow this repo to push them.

castrojo:
I've disabled the hwe build workflows in GitHub to prep. Have not made any code changes so we can easily turn them back on. Amazing work ya'll!

castrojo:
> You ask a good question. I've actually started wondering why we build main images, much less nvidia. At the moment, my answer is precedent.

I think if we would have started with bootc/dnf we would have not needed to make these. It's now pretty easy to just turn on the non-free rpmfusion/negativo stuff in a script using normal bash and dnf and have the entire product be standalone, like how Bluefin LTS does it.

Ideally I think the future of main is base-main and base-nvidia with the expectation that the consumer will just `dnf groupinstall @desktop-group` in their custom image. There's also value in deriving from silverblue and kinoite - in LTS we kind of assemble the desktop by hand, which also has it's plusses and minuses. Shrug.

p5:
How Renovate is currently configured:

Renovate will automatically update the digests (not tags) of all images inside image-versions.yaml via PRs.
These PRs will be automatically merged if all required status checks are successful.
A Renovate PR can bump as many or as few digests in one go.  So it will handle the case where akmods-nvidia fails to build, but the other akmods builds are successful just fine **as long as the existing build of akmods-nvidia has a compatible kernel**.

bsherman:
> How Renovate is currently configured:
> 
> Renovate will automatically update the digests (not tags) of all images inside image-versions.yaml via PRs. These PRs will be automatically merged if all required status checks are successful. A Renovate PR can bump as many or as few digests in one go. So it will handle the case where akmods-nvidia fails to build, but the other akmods builds are successful just fine **as long as the existing build of akmods-nvidia has a compatible kernel**.

Thank you! I think this confirms that I didn't break things with my change.

p5 (approved):

m2Giles (approved):
The only major thing I'm not a fan of is the use of `yq` but that was already being used.

But overall this looks good.

PR #879, opened 2025-05-16
feat: add tpm2 pkcs11 toolset

lethedata:
Enables TPM2 to be used as the PKCS#11 cryptographic token.

- tpm2-pkcs11 : enables TPM2 PKCS#11 functionality
- tpm2-pkcs11-tools : tools required to setup and configure it

This extended the usefulness of a device's TPM2 to do things such as:
 - Store sshd host keys
 - SSH authentication via TPM2
 - Use TPM2 as OpenSSL crypto engine
 - Store OpenVPN certs

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -52,6 +52,8 @@
                 "symlinks",
                 "tcpdump",
                 "tmux",
+                "tpm2-pkcs11",
+                "tpm2-pkcs11-tools",
                 "traceroute",
                 "vim",
                 "wireguard-tools",


lethedata:
While the `tpm2-pkcs11-tools` could be left out and installed via pip[^1], it seemed reasonable to include it as well as they are needed to truly utilize `tpm2-pkcs11` anyway.

[^1]: https://github.com/tpm2-software/tpm2-pkcs11/blob/master/docs/INSTALL.md#installing-tpm2_ptoolpy

Issue #249, opened 2023-06-18
Include Japanese and Chinese input in KDE images 


nonetrix:
### Describe the package

Currently out of the box it is impossible to type in Japanese or Chinese or any language that requires a IME(Input Method Editor), this is also the case on Fedora KDE and Fedora Kinoite to my knowledge as the dependencies are not installed. This is especially bad with a immutable system in my opinion. As such I think `fcitx5-anthy` or `fcitx5-mozc`, `fcitx5`, `fcitx-qt5`, `kcm-fcitx5`, `fcitx-gtk3` and other packages needed for Japanese support should be added, I wouldn't know what the ones needed for Chinese are however. This is a huge usability oversight in my opinion for many users including myself.

I have only seen this issue on the Kinoite version, however it might effect other spins as well. However, it might be useful to note this issue doesn't effect the GNOME version, it works fine there luckily for whatever reason :D 

## fcitx5-anthy
```
Available Packages
Name         : fcitx5-anthy
Version      : 5.0.14
Release      : 1.fc38
Architecture : x86_64
Size         : 219 k
Source       : fcitx5-anthy-5.0.14-1.fc38.src.rpm
Repository   : fedora
Summary      : Anthy Wrapper for Fcitx5
URL          : https://github.com/fcitx/fcitx5-anthy
License      : GPLv2+
Description  : Anthy Wrapper for Fcitx5
             : Ported from scim-anthy. Released under GPL2+.
```
## fcitx5-mozc
```
Available Packages
Name         : fcitx5-mozc
Version      : 2.17.2102.102.1
Release      : 24.20230320git242b4f7.fc38
Architecture : x86_64
Size         : 14 M
Source       : fcitx5-mozc-2.17.2102.102.1-24.20230320git242b4f7.fc38.src.rpm
Repository   : fedora
Summary      : A wrapper of mozc for fcitx5
URL          : https://github.com/fcitx/mozc
License      : BSD and ASL 2.0 and UCD and Public Domain and mecab-ipadic and LGPLv2+ and MS-PL
Description  : A wrapper of mozc for fcitx5.
```
## fcitx5
```
Available Packages
Name         : fcitx5
Version      : 5.0.23
Release      : 1.fc38
Architecture : i686
Size         : 1.8 M
Source       : fcitx5-5.0.23-1.fc38.src.rpm
Repository   : fedora
Summary      : Next generation of fcitx
URL          : https://github.com/fcitx/fcitx5
License      : LGPLv2+
[...]

PR #913, opened 2025-06-01
feat: add fcitx5 packages

inffy:
Adds fcitx5 packages to the main images.

Closes https://github.com/ublue-os/main/issues/249


ledif (approved):
Based on feedback in https://github.com/ublue-os/main/issues/249, these seems good to go.

Issue #927, opened 2025-06-07
Deprecate images not used by our products


castrojo:
Now that F42 is out it's time to clean out the tech debt. Since none of the core products use the wide variety of images created here I propose removing them. 

We should keep: 

main-base <-- empty solid base
silverblue-main <-- used by bluefin and bazzite
kinoite-main <-- used by aurora and bazzite

Since no core developers use anything else it makes sense to deprecate them, especially given that bootc is in cncf sandbox and multiple orgs are now springing up so we don't need to do this. I think something like 90 days and/or break/fix, whichever comes first?

Additionally we should trim down the "build your own!" stuff on the website but we can do that later. 

Only bluefin uses GTS so those can probably go asap.

tulilirockz:
Would we still build `-nvidia` images? None of the ublue downstreams are actually using `-nvidia` stuffs, just using the `nvidia-install.sh` script.

bsherman:
@tulilirockz 
> Would we still build `-nvidia` images? None of the ublue downstreams are actually using `-nvidia` stuffs, just using the `nvidia-install.sh` script.

We actually didn't discuss removing nvidia, though you make a valid point; it would be more consistent with our  statement here if we do, since the nvidia images are also not used in our downstream final images.

The reason we don't use it in those final images is largely the fact that we do kernel swaps in them, while in main we only build the default kernel, and thus nvidia images with the default kernel.


@castrojo @m2Giles @KyleGospo @ublue-os/approver 
should we remove nvidia builds? 
If so, I definitely think we should add a deprecation notice, like we did for the other builds we plan to remove in September, not do so immediately.

antheas:
I was also speaking with people in flock and a general theme was that you cannot use Nvidia on silver blue and kinoite, so I would keep them for this reason

bsherman:
> I know the Nvidia images are something that people use. @travier even has them mentioned in his readme and has recommended them.

> I was also speaking with people in flock and a general theme was that you cannot use Nvidia on silver blue and kinoite, so I would keep them for this reason

These are good points to be aware of, but also, still directly in conflict with us trying to streamline the build processes to only things we consume ourselves... it keeps getting delayed for many reasons, but when we finally add ARM builds, we instantly double _ALL_ of our builds so I've always wanted to keep the initial set of builds low for that reason... avoiding the extra builders and complexity.

I wonder if a suitable alternative is making the default example in the `image-template` repo literally build an nvidia image from main?

[...]

PR #935, opened 2025-06-08
chore: remove all images that are not being used by universal blue

tulilirockz:
Just doesnt remove `-nvidia` yet, need some discussion on that


bsherman:
I'll discuss the idea of removing nvidia on the issue where you asked the question https://github.com/ublue-os/main/issues/927#issuecomment-2953916066

But with regards to moving MORE of the code that supports other image builds, I intentionally left that code in place for interested parties.

PR #952, opened 2025-06-14
fix: Fix Nvidia Install Script and other various fixes

m2Giles:
changes: Add back cache mounts for /var/cache and /var/log.
fix: DRACUT xattr failing on local builds.
fix: nvidia-install script did not use config-manager for enabling/disabling repos causing fedora-multimedia to not actually get disabled.
fix: run bootc container lint in own run cmd to not pickup run mounts.

### DIFF
--- a/build_files/nvidia-install.sh
+++ b/build_files/nvidia-install.sh
@@ -3,20 +3,18 @@
 set -ouex pipefail
 
 RELEASE="$(rpm -E %fedora)"
-AKMODNV_PATH=${AKMODNV_PATH:-/tmp/akmods-rpms}
+: "${AKMODNV_PATH:=/tmp/akmods-rpms}"
 
 # this is only to aid in human understanding of any issues in CI
 find "${AKMODNV_PATH}"/
 
-if [[ ! $(command -v dnf5) ]]; then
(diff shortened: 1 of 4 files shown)


inffy (approved):

build_files/nvidia-install.sh
```diff
@@ -121,5 +117,5 @@ fi
 
 # re-enable negativo17-mutlimedia since we disabled it
 if [[ "${NEGATIVO17_MULT_PREV_ENABLED}" = "Y" ]]; then
-    sed -i '0,/enabled=0/{s/enabled=0/enabled=1/}' /etc/yum.repos.d/negativo17-fedora-multimedia.repo
+    dnf5 config-manager setopt fedora-multimedia.enabled=1
```
bsherman, line 120: Using setopt to enable this seems inconsistent with other places where we use --set-enabled 

Is there a reason you chose this specifically?

Comment applies to all usages in this PR

bsherman:
TIL thank you!

Will need to watch that especially close on build stuff which is trying to
handle both centos and Fedora.

bsherman (approved):

PR #983, opened 2025-06-24
fix(override): pin glibc to 2.41-5

m2Giles:
Addressing https://github.com/fedora-silverblue/issue-tracker/issues/660

Signed-off-by: m2 <69128853+m2Giles@users.noreply.github.com>


bsherman (approved):

p5 (approved):
LGTM

m2Giles:
Looks like this is only necessary on F42

p5:
Nvidia seems to depend on a modern version.
Maybe downgrade without the version pinning?

PR #1029, opened 2025-07-08
chore: build only necessary images

p5:


m2Giles (approved):
I like this a lot. I think inline makes sense here, but maybe we can make this more agnostic in the future to turn into an action.

inffy (approved):

PR #1155, opened 2025-08-13
chore: enable Fedora 43 builds

p5:
Disabled packages for F43:
- nvidia-container-toolkit (does not support RPM6 - https://github.com/NVIDIA/nvidia-container-toolkit/pull/1351)
- android-udev-rules (has some postinstall scripts that fail to call systemd)
- mesa-libxatracker (no F43 builds)
- pipewire-libs-extra (no F43 builds)

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -3,7 +3,6 @@
         "include": {
             "all": [
                 "alsa-firmware",
-                "android-udev-rules",
                 "apr",
                 "apr-util",
                 "distrobox",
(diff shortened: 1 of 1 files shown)


inffy (approved):

bsherman:
> Disabled packages for F43:
> 
> * nvidia-container-toolkit (does not support RPM6 - [Produce RPM packages with SHA256 header digests NVIDIA/nvidia-container-toolkit#1351](https://github.com/NVIDIA/nvidia-container-toolkit/pull/1351))
> * android-udev-rules (has some postinstall scripts that fail to call systemd)
> * mesa-libxatracker (no F43 builds)
> * pipewire-libs-extra (no F43 builds)

@p5 thank you for the clear call outs on what's missing.

PR #1224, opened 2025-08-27
fix(ci): reference upstream cosign public keys from main branch

p5:
The cosign public key has disappeared from the version-specific branches.  Switch to pulling them from main.


antheas (approved):
This really should be in our repo though. Like what is even the point of checking this if you pull just pull it from the repo that'd be compromised.

p5:
Will merge this after 1 more approval and bypass merge queue.  Builds will be ran on the Renovate PR.

inffy (approved):

PR #2039, opened 2026-01-30
chore: remove OpenCL-ICD-Loader workaround

### LINKED ISSUE #1749: can the OpenCL ICD bug workaround be dropped?
Re:

https://github.com/ublue-os/main/blob/28c12d7314d3c0ea9186d2ec1c3b859251f12b4d/build_files/install.sh#L26-L29

The linked bugzilla is closed as fixed and the fixing patch is https://src.fedoraproject.org/rpms/mesa/c/1940fd5d83011d24aaafc231c1e1e0069df79662.

renner0e:
This seems to have been resolved. `clinfo` tells me that everything still works so I think this is fine to remove.

Closes: https://github.com/ublue-os/main/issues/1749

### DIFF
--- a/build_files/install.sh
+++ b/build_files/install.sh
@@ -13,11 +13,6 @@ if ! rpm -q dnf5 >/dev/null; then
     rpm-ostree install dnf5 dnf5-plugins
 fi
 
-# mitigate upstream packaging bug: https://bugzilla.redhat.com/show_bug.cgi?id=2332429
-# swap the incorrectly installed OpenCL-ICD-Loader for ocl-icd, the expected package
-dnf5 -y swap --repo='fedora' \
-    OpenCL-ICD-Loader ocl-icd
-
 # Add COPRs
 dnf5 -y copr enable ublue-os/packages
 dnf5 -y copr enable ublue-os/staging


inffy (approved):

renner0e:
so to add a little bit more insight to this, we still need this for F42. I'm gonna draft this and this can be merged when F42 is dropped here as this is Fixed in F43+

bsherman:
Wow! @renner0e I completely dropped the ball on this PR!

I saw it today and realized I should finally make good on my promise to test on NVIDIA.

So, I tested this on Fedora 44 (silverblue-nvidia) with an NVIDIA GTX 1660 Ti (driver 595.71.05). I used a custom image which added `clinfo` and swapped `ocl-icd back` to `OpenCL-ICD-Loader` to simulate what this PR would do. I then compared `clinfo` before and after. 

**Good news**: the GPU itself works exactly the same either way. Same platform, same device info, same driver, same capabilities. Nothing about actually running OpenCL apps seemed to change.

**But I did find one difference.** It has to do with how the two ICD loaders behave when a program asks for "the platform" without saying which one it means (passing `NULL` instead of a specific platform ID).

- With today's `ocl-icd` (before this PR): if `NULL` is passed, it just assumes "well there's only one platform installed, use that one" and it works.
- With the Khronos loader (after this PR): passing `NULL` just fails outright with "No platform."

**What that means in practice**: based on checking with AI, because I'm not an OpenCL developer... apparently, programs aren't supposed to pass `NULL` like that — the spec expects apps to ask "what platforms exist?" first and then pick one explicitly. So any well-written, up-to-date OpenCL app shouldn't notice this change. The risk is with older or lazier OpenCL programs that skip that step and just assume there's one platform. Those could break after this merges — not because the GPU stopped working, but because the app didn't ask for it the "right" way and the new loader is stricter about that. That's more likely to bite something like an older scientific/rendering/transcoding tool than typical desktop use.

This shouldn't be a blocker, Fedora already made the decision, and arguably the Khronos loader is just being more spec-correct — but it's a change from what ships today, so figured it's worth commenting on rather than just saying "clinfo looks fine."

bsherman (approved):

bsherman:
Also, this is going to fail merge checks since nvidia isn't building right now, but can be force merged if desired.

renner0e:
Thanks for looking more deeply into this

PR #2261, opened 2026-03-28
remove Fedora 42 & GTS builds

nikodunk:
It appears to me F42 & GTS builds could be removed to lighten the load on the runners.


m2Giles:
Did we deprecate GTS for main as well?

I was tracking that Bluefin deprecated their GTS images.

PR #2436, opened 2026-05-09
chore: remove f42 remainings

Lumaeris:
No reason to pull in F42 updates when they're no longer in use here :P

Adding F45 to image-version is too early rn


bsherman (approved):
LGTM

inffy (approved):

PR #2527, opened 2026-06-05
fix(nvidia): update nvidia packages for v610 driver

inffy:
Noticed that our main-nvidia images have been failing for quite some time because of the nvidia 610 driver update that changed some packages.


grulja:
Hi, why this was closed? It looks nvidia images are still not being produced.

